PR #2, opened 2019-03-06
Feature/impl storage

pyakushin:


src/storage.rs
```diff
@@ -14,7 +24,26 @@ impl Storage {
...
+        self.prepare_work_dir()?;
+        let wd = Path::new(&self.config.work_dir);
+        let files_in_work_dir: Vec<_> = fs::read_dir(wd)?
+            .map(|entry| entry.unwrap().file_name())
+            .collect();
+        if files_in_work_dir.is_empty() {
+            println!("working dir is empty, start new storage");
+            self.init_new().unwrap();
+        } else {
+            println!("working dir contains files, try init existing");
```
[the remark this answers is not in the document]
pyakushin: We have not yet decided on a logging system, so the `println` is temporarily
I think to use `log` crate macro as facade, so in order to produce log output executables  have to use a logger implementation compatible with the facade.
I can attach, e.g. `env_logger` only for tests

src/storage.rs
```diff
@@ -40,12 +69,70 @@ impl Storage {
...
+                _ => return Err(e),
+            }
+        }
+        println!("create dir all");
+        fs::create_dir_all(wd)?;
+        Ok(())
+    }
+
+    fn init_new(&mut self) -> Result<(), ()> {
+        self.opened_blob = Box::new(Some(Default::default()));
```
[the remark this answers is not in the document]
agend: I have thought a little bit more: let use name: active blob in naming

src/storage.rs
```diff
@@ -14,20 +24,44 @@ impl Storage {
...
     /// ```
-    pub fn init(&mut self) -> Result<(), ()> {
+    ///
+    /// Storage works in dir provided to builder. If dir not exist,
+    /// creates it, otherwise tries init dir as existing storage.
+    pub fn init(&mut self) -> io::Result<()> {
+        // @TODO implement work dir validation
+        self.prepare_work_dir()?;
+        let wd = Path::new(&self.config.work_dir);
+        let files_in_work_dir: Vec<_> = fs::read_dir(wd)?
```
agend: doing twice read_dir

src/storage.rs
```diff
@@ -36,16 +70,99 @@ impl Storage {
...
+        if let Err(e) = fs::read_dir(wd) {
+            match e.kind() {
+                ErrorKind::NotFound => {
+                    error!("\"{}\" not found", self.config.work_dir);
+                }
+                _ => return Err(e),
+            }
+        }
+        debug!(
+            "create work dir recursively: {}",
```
agend: create->creating

src/storage.rs
agend: we can return file list from prepare_work_dir

src/storage.rs
```diff
@@ -36,16 +70,99 @@ impl Storage {
...
+    // @TODO specify more useful error type
+    fn init_from_existing(&mut self, files: Vec<DirEntry>) -> Result<(), ()> {
+        self.blobs = files
+            .iter()
+            .filter_map(|entry| {
+                // @TODO implement more file validations
+                if entry.metadata().ok()?.is_file() {
+                    Some(Blob::from_file(entry.path()).ok()?)
+                } else {
+                    None
```
agend: we can log that we skipping this file

src/storage.rs
```diff
@@ -36,16 +70,99 @@ impl Storage {
...
+                }
+            })
+            .collect();
+        if let Some(blob_index) = self.search_for_opened_blob() {
+            self.opened_blob = Box::new(Some(self.blobs.remove(blob_index)));
```
agend: we can use `blobs` instead of opened_blobs (as we would have opened blobs and active one)

src/storage.rs
```diff
@@ -36,16 +70,99 @@ impl Storage {
...
+            Ok(())
+        } else {
+            self.init_opened_blob()
+        }
+    }
+
+    fn search_for_opened_blob(&self) -> Option<usize> {
```
agend: for what purpose this function?

src/storage.rs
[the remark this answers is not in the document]
pyakushin: Replace `println!` with macro from `log` crate

src/storage.rs
```diff
@@ -40,12 +69,70 @@ impl Storage {
...
+    }
+
+    fn init_new(&mut self) -> Result<(), ()> {
+        self.opened_blob = Box::new(Some(Default::default()));
+        Ok(())
+    }
+
+    fn init_from_existing(&mut self) -> Result<(), ()> {
+        let wd = Path::new(&self.config.work_dir);
+        let entries: Vec<_> = fs::read_dir(wd)
```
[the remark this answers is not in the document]
pyakushin: Fixed. Pass entries here from `init()`

src/storage.rs
```diff
@@ -40,12 +69,70 @@ impl Storage {
...
+            .map_err(|e| eprintln!("{}", e))?
+            .map(std::result::Result::unwrap)
+            .collect();
+        self.blobs = entries
```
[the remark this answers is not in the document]
pyakushin: Added TODO

src/storage.rs
```diff
@@ -40,12 +69,70 @@ impl Storage {
...
+            .iter()
+            .filter_map(|entry| {
+                if entry.metadata().ok()?.is_file() {
+                    Some(Blob::from_file(entry.path()).ok()?)
+                } else {
+                    None
+                }
+            })
+            .collect();
+        self.init_new()?;
```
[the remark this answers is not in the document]
pyakushin: Done. Find first opened blob

src/storage.rs
```diff
@@ -36,16 +70,99 @@ impl Storage {
...
+    }
+
+    // @TODO specify more useful error type
+    fn init_new(&mut self) -> Result<(), ()> {
+        self.init_opened_blob()
+    }
+
+    // @TODO specify more useful error type
+    fn init_from_existing(&mut self, files: Vec<DirEntry>) -> Result<(), ()> {
+        self.blobs = files
```
agend: enumerating files from disk could have unpredictable order. we need to extract some id we will embed into file name example: pblob-1, pblob-2 and then sort it by this name in code.

src/storage.rs
pyakushin: Fixed. Check existence with [Path::exists(_)](https://doc.rust-lang.org/std/path/struct.Path.html#method.read_dir)

src/storage.rs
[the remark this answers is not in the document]
pyakushin: Fixed

src/storage.rs
pyakushin: Fixed

src/storage.rs
agend: We can't add data to closed blobs. But we can update metadata - ex: delete or update in future

src/storage.rs
pyakushin: I enumerating `self.blobs: Vec<Blobs>` and indexing the same `self.blobs`. Indexing blobs by id from their name marked as TODO

src/storage.rs
pyakushin: Yes, I mean that Storage already has two fields for 1 active and N closed blobs

src/blob.rs
```diff
@@ -15,6 +17,17 @@ struct Header {
...
+        // @TODO implement
+        unimplemented!()
+    }
+
+    pub fn is_active(&self) -> bool {
+        unimplemented!()
+    }
+}
+
 #[test]
```
ikopylov: Is is ok to write tests in the file with implementation?

tests/tests.rs
```diff
@@ -1,28 +1,30 @@
 use pearl::Builder;
 
 #[test]
-fn test_storage_init() {
-    let storage_builder = Builder::new().work_dir("/tmp/");
-    let _storage = storage_builder.build();
+fn test_storage_init_new() {
+    let storage_builder = Builder::new().work_dir("/tmp/pearl/");
+    let mut storage = storage_builder.build();
+    assert!(storage.init().map_err(|e| eprintln!("{}", e)).is_ok());
```
DenisBabarykin: It is not good to use asserts in both debug and release modes. In release mode you should handle all error situations gracefully and of course make logs about it. For debug stage of development you better use [debug asserts](https://doc.rust-lang.org/std/macro.debug_assert.html) that will not be included in code in release mode. But I recommend you to not use runtime asserts at all and use only [static asserts](https://crates.io/crates/static_assertions) which works only in compile time.

src/storage.rs
```diff
@@ -36,16 +70,94 @@ impl Storage {
...
+        } else {
+            self.init_active_blob()
+        }
+    }
+
+    fn search_for_active_blob(&self) -> Option<usize> {
+        // @TODO Search for last active blob by index extracted from file name
+        let mut active_blob_id = None;
+        self.blobs.iter().enumerate().find(|(i, blob)| {
+            let res = blob.is_active();
```
ikopylov: Do the blob need to know if he is active or not? I suggest to determine the active blob as the last one in the directory.

src/storage.rs
```diff
@@ -36,16 +70,94 @@ impl Storage {
     /// # Description
     /// Closes all file descriptors
     /// # Examples
+
+    // @TODO specify more useful error type
     pub fn close(&mut self) -> Result<(), ()> {
```
ikopylov: Why was the explicit `close` method created? Probably its implementation should be part of the destructor.

tests/tests.rs
DenisBabarykin: Oh, I didn't notice, that it is tests, sorry :)

src/blob.rs
pyakushin: You puts unit tests in each file in _src_ , with `#[cfg(test)]` annotation
Integration tests are entirely external to library and locates in _tests_ dir at the top level of project
Src: [Rust Book Test organization](https://doc.rust-lang.org/book/ch11-03-test-organization.html)

tests/tests.rs
pyakushin: Thanks for the review) 
In addition, Rust `assert`s rarely used in safe code, in general in unsafe code and tests.

src/storage.rs
pyakushin: Fair point about who knows about the state of a blob, 'cause `Storage` contains configuration and can check the state.
Second, if the last blob is full, a new one must be initialized

src/storage.rs
pyakushin: It created for case when you want to close all file descriptors without dropping the storage. 
(can be removed)

src/blob.rs
ikopylov: >  The convention is to create a module named tests in each file to contain the test functions and to annotate the module with cfg(test).

According to Rust Book you should define a separate module.

src/storage.rs
ikopylov: Why is it even needed? When all descriptors are closed, Storage object can't be used anyway.

PR #6, opened 2019-03-07
Finalize storage initialization

pyakushin:
Current status:
- Storage builder requires key type on `build()`
- Finished config parameters validation and unit tests for:
  - `work_dir`
  -  `max_blobs_num`

### DIFF
--- a/src/blob.rs
+++ b/src/blob.rs
@@ -17,10 +17,10 @@ struct Header {
     flags: u64,
 }
 
-impl<T> Blob<T> {
+impl<T> Blob<T> where T: Default {
     pub fn from_file<P: AsRef<Path>>(_path: P) -> Result<Self, ()> {
         // @TODO implement
-        unimplemented!()
+        Ok(Self::default())
     }
 }
 
--- a/src/blob.rs
+++ b/src/blob.rs
@@ -1,10 +1,10 @@
 use std::path::Path;
 
-/// A `Blob` struct for performing of database
-/// # Examples
+/// A `Blob` struct for performing of database,
 #[derive(Debug, Default)]
-pub struct Blob {
+pub struct Blob<T> {
     header: Header,
+    records: Vec<T>, // @TODO needs verification, created to yield generic T up
 }
 
 /// # Description
@@ -17,7 +17,7 @@ struct Header {
     flags: u64,
 }
 
-impl Blob {
+impl<T> Blob<T> {
     pub fn from_file<P: AsRef<Path>>(_path: P) -> Result<Self, ()> {
         // @TODO implement
         unimplemented!()
@@ -30,8 +30,9 @@ mod tests {
 
     #[test]
     fn test_blob_new() {
-        let _ = Blob {
+        let _b: Blob<u32> = Blob {
             header: Default::default(),
+            records: Vec::new(),
         };
     }
 }
--- a/src/storage.rs
+++ b/src/storage.rs
@@ -1,16 +1,18 @@
 use std::{
-    fs::{self, DirEntry},
+    fs::{self, DirEntry, OpenOptions},
     io,
     path::{Path, PathBuf},
 };
 
 use crate::blob::Blob;
 
+const TEST_FILE_PATH: &str = "test_file";
+
 /// # Description
 /// Used to create a storage, configure it and manage
 /// `T` - type of storage key, must be [Sized](https://doc.rust-lang.org/std/marker/trait.Sized.html)
 /// # Examples
-/// ```
+/// ```no-run
 /// use pearl::{Storage, Builder};
 ///
 /// let mut storage = Builder::new().build::<u32>();
@@ -31,7 +33,7 @@ where
     /// Creates a new instance of a storage with u32 key
     /// # Examples
     ///
-    /// ```
+    /// ```no-run
     /// use pearl::{Storage, Builder};
     ///
     /// let mut storage = Builder::new().build::<u32>();
@@ -93,7 +95,7 @@ where
     /// # Description
     /// Blobs count contains closed blobs and one active, if is some.
     /// # Examples
-    /// ```
+    /// ```no-run
     /// use pearl::Builder;
     /// // key type f64
     /// let mut storage = Builder::new().work_dir("/tmp/pearl/").build::<f64>();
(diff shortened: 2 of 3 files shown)


src/storage.rs
```diff
@@ -168,18 +184,62 @@ impl<'a> Builder {
...
+        info!("work dir set: {}", path.display());
+        self.config.work_dir = Some(path);
+        self
+    }
+
+    /// # Description
+    /// Sets storage max blob files num
+    /// Must be greater than zero
+    /// Limited by `u32::MAX`
+    pub fn max_blobs_num<U: Into<u32>>(mut self, max_blobs_num: U) -> Self {
```
agend: ?

src/storage.rs
```diff
@@ -168,18 +184,62 @@ impl<'a> Builder {
...
+        debug!("check is path readonly");
+        debug!("check is path a dir");
+        if permissions.readonly() {
+            error!("{} is in readonly mode", path.display());
+            panic!("work dir must have write permission");
+        } else if !metadata.is_dir() {
+            error!("{} is not a directory", path.display());
+            panic!("work dir must be a directory");
+        }
+        info!("work dir set: {}", path.display());
```
agend: Check by creating file in folder

src/storage.rs
```diff
@@ -168,18 +184,62 @@ impl<'a> Builder {
...
+        let metadata = path
+            .metadata()
+            .map_err(|e| error!("metadata load failed: {}", e))
+            .unwrap();
+        let permissions = metadata.permissions();
+        debug!("check is path readonly");
+        debug!("check is path a dir");
+        if permissions.readonly() {
+            error!("{} is in readonly mode", path.display());
+            panic!("work dir must have write permission");
```
agend: should be error not panic

src/blob.rs
```diff
@@ -1,10 +1,10 @@
 use std::path::Path;
 
-/// A `Blob` struct for performing of database
-/// # Examples
+/// A `Blob` struct for performing of database,
 #[derive(Debug, Default)]
-pub struct Blob {
+pub struct Blob<T> {
```
ikopylov: IMHO, It is better to name type parameter as `TKey`

src/storage.rs
```diff
@@ -1,36 +1,49 @@
...
-    blobs: Vec<Blob>,
+    active_blob: Box<Option<Blob<T>>>,
+    blobs: Vec<Blob<T>>,
 }
 
-impl Storage {
-    /// Creates a new instance of a storage
+impl<T> Storage<T>
+where
+    T: Default,
```
ikopylov: I think we should define trait for storage key to allow extensibility in the future

src/storage.rs
ikopylov: +1

We should specify a capacity of single blob (not the maximum number of blobs in storage)

src/storage.rs
```diff
@@ -113,11 +115,19 @@ where
         let path = Path::new(self.config.work_dir.as_ref().unwrap()); // @TODO handle unwrap explicitly
         if !path.exists() {
             debug!("creating work dir recursively: {}", path.display());
-            fs::create_dir_all(path)
+            fs::create_dir_all(path)?;
         } else {
             debug!("work dir exists: {}", path.display());
-            Ok(())
         }
+        let test_path = path.join(TEST_FILE_PATH);
```
agend: maybe to use `.lock` file - as mark of open directory by storage. could be helpful for debuging some case + we can store some meta there. Behavior - we can open/create this file and close on storage close.

src/storage.rs
```diff
@@ -194,51 +204,29 @@ impl<'a> Builder {
...
-    /// Limited by `u32::MAX`
-    pub fn max_blobs_num<U: Into<u32>>(mut self, max_blobs_num: U) -> Self {
-        let mbn = max_blobs_num.into();
-        debug!("cmp {} with 0", mbn);
-        if mbn > 0 {
-            self.config.max_blobs_num = Some(mbn);
-            info!("maximum blobs number set to: {}", mbn);
+    pub fn max_blob_size<U: Into<usize>>(mut self, max_blob_size: U) -> Self {
+        let mbs = max_blob_size.into();
+        debug!("cmp {} with 0", mbs);
```
agend: Why do we need this?

src/storage.rs
```diff
@@ -194,51 +204,29 @@ impl<'a> Builder {
...
+        if mbs > 0 {
+            self.config.max_blob_size = Some(mbs);
+            info!("maximum blob size set to: {}", mbs);
         } else {
-            error!("zero blobs num is useless, not set");
+            error!("zero size blobs is useless, not set");
```
agend: We can consider `0` as - use single blob file always (could be some use case)

src/storage.rs
```diff
@@ -261,22 +248,32 @@ mod tests {
     use super::*;
     use std::fs::*;
 
-    const RO_DIR_NAME: &str = "/tmp/pearl_test/readonly_folder/";
-    const FILE_NAME: &str = "/tmp/pearl_test/work_dir.file";
+    const RO_DIR_NAME: &str = "/tmp/pearl_test_readonly/";
```
agend: We can use https://doc.rust-lang.org/std/env/fn.temp_dir.html

src/storage.rs
ikopylov: I am not agree. Zero size of the blob is definitely invalid value. For the unlimited size of the blob it is easier to set some large value for this field (obviously, there is no storage system that can contain file of size = 2^64 bytes)

src/storage.rs
pyakushin: added `lock.file`

src/storage.rs
pyakushin: If you will change Key trait, it'll break backward compatibility. So it is the same as adding new trait if we will need it.

src/storage.rs
pyakushin: Left as is

src/storage.rs
```diff
@@ -225,24 +233,74 @@ impl<'a> Builder {
...
+        }
+        self
+    }
+
+    /// # Description
+    /// Sets blob file name pattern, e.g. if pattern set to `hellopearl`,
+    /// files will be named as `hellopearl.[N].blob`.
+    /// Where N - index number of file
+    /// Must be not empty
+    pub fn blob_file_name_pattern<U: Into<String>>(mut self, blob_file_name_pattern: U) -> Self {
```
agend: pattern->prefix

PR #7, opened 2019-03-13
Feature/storage read write init

pyakushin:


src/blob.rs
```diff
@@ -17,11 +17,56 @@ struct Header {
...
+    // @TODO more useful result
     pub fn from_file<P: AsRef<Path>>(_path: P) -> Result<Self, ()> {
         // @TODO implement
         Ok(Self::default())
     }
+
+    /// # Description
+    /// Writes given slice to file, returns number of written bytes
+    // @TODO more useful result
+    pub fn write(&mut self, _key: &K, _value: &[u8]) -> Result<usize, ()>  {
```
agend: We should write some struct and not separate key and value (value is bad name for this any way).

agend: I don't think we are interested that exact number of bytes has been writen to the disk. may be err in return will be enough.

src/blob.rs
```diff
@@ -17,11 +17,56 @@ struct Header {
...
+    // @TODO more useful result
+    pub fn size(&self) -> Result<usize, ()> {
+        // @TODO implement
+        Ok(0usize)
+    }
+
+    /// # Description
+    /// Returns number of records in current blob
+    // @TODO more useful result
+    pub fn len(&self) -> Result<usize, ()> {
```
agend: len -> count

src/blob.rs
ikopylov: IMHO, `key` and `value` should be passed as separate arguments, but I agree that it is better to make `value` a structure.

src/blob.rs
```diff
@@ -17,11 +17,56 @@ struct Header {
...
+    // @TODO more useful result
+    pub fn read(&self, _key: &K) -> Result<Vec<u8>, ()>  {
+        // @TODO implement
+        Ok(Vec::new())
+    }
+
+    /// # Description
+    /// Closes current blob and sets blob file in readonly mode
+    // @TODO more useful result
+    pub fn close(&mut self) -> Result<(), ()> {
```
ikopylov: Probably, this method is redundant.

src/storage.rs
```diff
@@ -21,8 +21,9 @@ const LOCK_FILE: &str = "pearl.lock";
 ///
 #[derive(Debug)]
 pub struct Storage<K> {
+    initialized: Option<()>,
     config: Config,
-    active_blob: Box<Option<Blob<K>>>,
+    active_blob: Option<Box<Blob<K>>>,
```
ikopylov: This field is accessed from multiple threads thus it is required to use synchronization or memory barrier. In current shape it is not a pointer, so all access should happen inside critical section. If you can make it a pointer size type, then `volatile` should be enough to prevent torn read.

src/storage.rs
```diff
@@ -61,26 +62,58 @@ where
...
-        // @TODO implement
-        Ok(())
+    pub fn write(&mut self, key: K, value: &[u8]) -> Result<usize, ()> {
+        self.initialized.ok_or(())?;
+        // @TODO process unwrap explicitly
+        if self.active_blob.as_ref().unwrap().size()? + value.len()
+            > self.config.max_blob_size.unwrap()
+            || self.active_blob.as_ref().unwrap().len()? >= self.config.max_data_in_blob.unwrap()
+        {
+            let new_active = Box::new(Default::default());
```
ikopylov: Critical section is required to prevent creation of multiple blobs from concurrent threads

src/storage.rs
```diff
@@ -61,26 +62,58 @@ where
...
     // @TODO specify more useful error type
-    pub fn write(&mut self) -> Result<(), ()> {
-        // @TODO implement
-        Ok(())
+    pub fn write(&mut self, key: K, value: &[u8]) -> Result<usize, ()> {
+        self.initialized.ok_or(())?;
+        // @TODO process unwrap explicitly
+        if self.active_blob.as_ref().unwrap().size()? + value.len()
+            > self.config.max_blob_size.unwrap()
+            || self.active_blob.as_ref().unwrap().len()? >= self.config.max_data_in_blob.unwrap()
```
ikopylov: `active_blob` should be read inside critical section to prevent torn read. If you make it a pointer then critical section can be avoided.

src/storage.rs
```diff
@@ -61,26 +62,58 @@ where
...
+        {
+            let new_active = Box::new(Default::default());
+            // @TODO process unwrap explicitly
+            let mut old_active = self.active_blob.replace(new_active).unwrap();
+            old_active.close()?;
+            self.blobs.push(*old_active);
+        }
+        // @TODO process unwrap explicitly
+        self.active_blob.as_mut().unwrap().write(&key, &value)
```
ikopylov: `active_blob` should be read once as it can be changed in any time!

src/storage.rs
```diff
@@ -61,26 +62,58 @@ where
...
+    /// records with matching key, returns `Err(_)`
 
     // @TODO specify more useful error type
-    pub fn read(&self) -> Result<(), ()> {
-        // @TODO implement
-        Ok(())
+    pub fn read(&self, key: &K) -> Result<Vec<u8>, ()> {
+        // @TODO match error in map_err
+        if let Ok(data) = self
+            .active_blob
```
ikopylov: Synchronization once again!

src/storage.rs
```diff
@@ -61,26 +62,58 @@ where
...
+            .as_ref()
+            .ok_or(())?
+            .read(key)
+            .map_err(|_e| debug!("no records in active blob"))
+        {
+            Ok(data)
+        } else if let Some(data) = self.blobs.iter().find_map(|blob| {
```
ikopylov: Synchronization is required when accessing `blobs`. You can avoid it if you make `blobs` immutable and pointer sized.

src/storage.rs
```diff
@@ -21,8 +21,9 @@ const LOCK_FILE: &str = "pearl.lock";
 ///
 #[derive(Debug)]
 pub struct Storage<K> {
+    initialized: Option<()>,
     config: Config,
-    active_blob: Box<Option<Blob<K>>>,
+    active_blob: Option<Box<Blob<K>>>,
     blobs: Vec<Blob<K>>,
```
ikopylov: It is better to define `blobs` as immutable array. When a new blob added you just reallocate the array with additional space for that blob. In this way you can access `blobs` for read without synchronization.

src/storage.rs
agend: need to think about good rust synchronization mechanism here. Mutex or RwLock (https://doc.rust-lang.org/std/sync/struct.RwLock.html) as we can read independently (there is a syscall pread in nix which can read with offset without need to seek operation https://docs.rs/nix/0.13.0/nix/sys/uio/fn.pread.html)

ikopylov: https://docs.rs/crossbeam/*/crossbeam/atomic/struct.AtomicCell.html

src/blob.rs
pyakushin: Updated: `write()` takes `Record` struct, which contains key

src/storage.rs
pyakushin: Syncronization and concurrency will be included in one of the followin issues

src/blob.rs
agend: @ikopylov let go with record. to avoid data duplication

src/storage.rs
agend: it will be in #8

src/storage.rs
pyakushin: Impossible in current realization

src/storage.rs
pyakushin: In this case I can clone `Vec<_>` on read

src/blob.rs
pyakushin: It could be used for, e.g. `flush()`

src/storage.rs
pyakushin: Rusts `Box<_>` is a pointer type, doc: [Box](https://doc.rust-lang.org/std/boxed/struct.Box.html)

PR #9, opened 2019-03-15
Feature/blob init

pyakushin:


src/blob.rs
```diff
@@ -1,12 +1,55 @@
...
+};
 
 use crate::record::Record;
 
 /// A `Blob` struct for performing of database,
 #[derive(Debug, Default)]
-pub struct Blob<T> {
+pub struct Blob<K>
+where
+    K: Send,
```
ikopylov: Maybe `Default + Send`?
And probably it's time to create a trait for key that inherits from all required traits (currently `Default` and `Send`)

src/blob.rs
```diff
@@ -1,12 +1,55 @@
...
+{
     header: Header,
-    records: Vec<T>, // @TODO needs verification, created to yield generic T up
+    records: Vec<K>, // @TODO needs verification, created to yield generic T up
```
ikopylov: What kind of data this field is supposed to store? All keys? If so, why is it needed at all?

src/blob.rs
```diff
@@ -21,14 +64,33 @@ struct Header {
 
 impl<K> Blob<K>
 where
-    K: Default,
+    K: Default + Send,
 {
+    /// Creates new blob file or openes existing and truncates it
+    pub fn open_new<P: AsRef<Path> + Send + 'static>(path: P) -> BlobOpenFuture<K, P> {
```
ikopylov: Why `path` should be static? Obviously it will be generated at runtime by `Storage`

ikopylov: Since you defined `open` method, you shoud track the state of the `Blob` object. Currently it has two internal states: `Created` and `Opened`. If the `Blob` object in `Created` state then it should throw exception in `read` and `write` methods.

src/storage.rs
```diff
@@ -161,9 +176,24 @@ where
     }
 
     // @TODO specify more useful error type
-    #[inline]
     fn init_active_blob(&mut self) -> Result<(), ()> {
-        self.active_blob = Some(Box::new(Default::default()));
+        let new_blob_name = format!(
+            "{}.{}.{}",
+            self.config.blob_file_name_prefix.as_ref().ok_or(())?,
+            self.blobs.len(),
```
ikopylov: This is incorrect. Blob number should be calculated as the maximum from existing plus 1. If the old blob is deleted, the numbers will begin to conflict.

src/storage.rs
```diff
@@ -174,26 +204,36 @@ where
...
-            .filter_map(|entry| {
-                // @TODO implement more file validations
-                if entry.metadata().ok()?.is_file() {
-                    Some(Blob::from_file(entry.path()).ok()?)
+            .map(DirEntry::path)
+            .filter(|path| path.is_file())
+            .filter_map(|path| {
+                let blob = runtime.block_on(Blob::from_file(path.clone())).unwrap();
+                let stem = path.file_stem().unwrap().to_str().unwrap();
+                let index_str = stem.split('.').last()?;
```
ikopylov: Create 2 separate methods: one to genenerate the Blob file name and another to parse it. It is also a good idea to create a struct that stores a components of the Blob file name

src/storage.rs
```diff
@@ -161,9 +176,24 @@ where
...
+        );
+        let path = self
+            .config
+            .work_dir
+            .as_ref()
+            .ok_or(())?
+            .join(&new_blob_name);
+        let mut runtime = tokio::runtime::Runtime::new().unwrap();
+        // @TODO remove block on
+        let blob = runtime.block_on(Blob::open_new(path)).unwrap();
```
ikopylov: Why did you created an asynchronous `open` method and now just block on it? You should make the whole code asynchronous or make the `open` method synchronous.

src/storage.rs
```diff
@@ -174,26 +204,36 @@ where
...
+        let temp_blobs = files
             .iter()
-            .filter_map(|entry| {
-                // @TODO implement more file validations
-                if entry.metadata().ok()?.is_file() {
-                    Some(Blob::from_file(entry.path()).ok()?)
+            .map(DirEntry::path)
+            .filter(|path| path.is_file())
+            .filter_map(|path| {
+                let blob = runtime.block_on(Blob::from_file(path.clone())).unwrap();
```
ikopylov: Opening a BLOB should happen only after you have successfully parsed its file name!

src/storage.rs
```diff
@@ -174,26 +204,36 @@ where
...
+                let stem = path.file_stem().unwrap().to_str().unwrap();
+                let index_str = stem.split('.').last()?;
+                let id = index_str.parse().ok()?;
+                if let Some(ref mut i) = max_blob_file_index {
```
ikopylov: The code becomes hard to read. It is better to select active BLOB after you load all blobs into list by scanning that list.

src/blob.rs
```diff
@@ -19,68 +28,154 @@ struct Header {
...
+    buf: Option<Vec<u8>>,
+    current_offset: u64,
+}
+
+impl Future for WriteFuture {
+    type Output = Result<()>;
+
+    fn poll(mut self: Pin<&mut Self>, _waker: &Waker) -> Poll<Self::Output> {
+        let buf = self.buf.take().unwrap();
+        self.file.write_at(&buf, self.current_offset).unwrap();
```
agend: this is bad as it blocking. we can use waker to monitor fd. I have created #10

src/blob.rs
```diff
@@ -19,68 +28,154 @@ struct Header {
...
+    pub fn open_new(path: PathBuf) -> Result<Self> {
+        Ok(Self {
+            header: Default::default(),
+            file: Some(
+                fs::OpenOptions::new()
+                    .create_new(true)
+                    .append(true)
+                    .read(true)
+                    .open(&path)
+                    .map_err(Error::OpenNew)?,
```
agend: We need to know what was the root of error - pass string with description

src/blob.rs
```diff
@@ -19,68 +28,154 @@ struct Header {
...
+        let (offset, size) = self.locate(&key)?;
+        Some(ReadFuture {
+            f: self.file.as_ref()?.try_clone().ok()?,
+            k: key.as_ref().to_vec(),
+            offset,
+            size,
+        })
+    }
+
+    fn locate<K>(&self, key: &K) -> Option<(u64, u64)>
```
agend: may be structure? Now it's hard to understand

src/record.rs
```diff
@@ -20,58 +21,149 @@ where
...
     /// # Description
-    /// Returns number of bytes, struct `Record` uses on disk
-    pub fn size(&self) -> usize {
+    pub fn header_key_len(&self) -> u64 {
+        self.header.key_len
+    }
+
+    /// # Description
+    pub fn header_data_len(&self) -> u64 {
+        self.header.size - HEADER_UNALIGNED_SIZE as u64 - self.header.key_len
```
agend: may be we should store `data_size` as header size and key length should come from blob's header. Why waste space to store same value all the time in each record.

src/storage.rs
```diff
@@ -155,69 +163,115 @@ where
...
-        self.active_blob = Some(Box::new(Default::default()));
+    fn init_active_blob(&mut self) -> Result<()> {
+        let new_blob_name = format!(
+            "{}.{}.{}",
+            self.config
+                .blob_file_name_prefix
+                .as_ref()
+                .ok_or(Error::WrongConfig)?,
+            self.blobs.len(),
+            BLOB_FILE_ETENSION
```
agend: typo: `BLOB_FILE_ETENSION` -> `BLOB_FILE_EXTENSION `

src/record.rs
```diff
@@ -20,58 +21,149 @@ where
...
+    pub fn header_size() -> usize {
+        HEADER_UNALIGNED_SIZE
+    }
+}
+
+impl Record {
+    /// # Description
+    /// Serialize header to `Vec<u8>` bytes
+    pub fn to_raw(&self) -> Vec<u8> {
+        let mut buf = self.header.to_raw();
```
agend: Memory reallocation will take place. We should calculate buffer size and the fill it in with data

src/record.rs
```diff
@@ -20,58 +21,149 @@ where
...
+        // @TODO Header validation
+        let mut record = Self::new();
+        if buf.len() < HEADER_UNALIGNED_SIZE {
+            None
+        } else {
+            record.header = Header::from_raw(buf);
+            record.key = buf
+                [HEADER_UNALIGNED_SIZE..HEADER_UNALIGNED_SIZE + record.header.key_len as usize]
+                .to_vec();
+            record.data = buf[HEADER_UNALIGNED_SIZE + record.header.key_len as usize..].to_vec();
```
agend: Can we avoid to use `HEADER_UNALIGNED_SIZE` we have decided to go field by field serializing and deserializing?

src/blob.rs
```diff
@@ -1,12 +1,21 @@
...
+type Result<T> = std::result::Result<T, Error>;
 
 /// A `Blob` struct for performing of database,
 #[derive(Debug, Default)]
-pub struct Blob<T> {
+pub struct Blob {
+    index: BTreeMap<Vec<u8>, record::Header>,
     header: Header,
-    records: Vec<T>, // @TODO needs verification, created to yield generic T up
+    file: Option<fs::File>,
```
agend: Why `file` is `Option`?

src/blob.rs
pyakushin: It is outdated now

src/blob.rs
pyakushin: In fahrenheit implemented only AsyncTcpStream/AsyncTcpListener. So I have to write implementation for files by myself. Looks like the new pull request

src/blob.rs
pyakushin: [```Error::OpenNew```](https://github.com/qoollo/pearl/blob/feature/blob_init/src/blob.rs#L177) is a closure that wraps [```io::Result<T>```](https://doc.rust-lang.org/std/io/type.Result.html)

src/storage.rs
pyakushin: Done

src/storage.rs
pyakushin: Done

src/storage.rs
pyakushin: Outdated

src/storage.rs
pyakushin: outdated

src/record.rs
pyakushin: Updated with ```bincode::serialized_size(_)```

src/storage.rs
pyakushin: fixed

src/record.rs
pyakushin: [Same](https://github.com/qoollo/pearl/pull/9#discussion_r270936322)

src/blob.rs
pyakushin: removed

src/blob.rs
```diff
@@ -1,86 +1,258 @@
...
+        Box::new(self)
+    }
+
+    pub fn from_file(path: PathBuf) -> Result<Self> {
+        let file = fs::OpenOptions::new()
+            .create(false)
+            .append(true)
+            .read(true)
+            .open(&path)
+            .map_err(Error::FromFile)?;
```
ikopylov: There should be a method that scans the file and checks it for correctness. The check for magic byte should always be executed. The full scan for checksum correctness should be optional.

src/blob.rs
```diff
@@ -1,86 +1,258 @@
...
+const BLOB_MAGIC_BYTE: u64 = 0xdeaf_abcd;
+
+type Result<T> = std::result::Result<T, Error>;
 
 /// A `Blob` struct for performing of database,
-#[derive(Debug, Default)]
-pub struct Blob<T> {
+#[derive(Debug)]
+pub struct Blob {
+    index: BTreeMap<Vec<u8>, record::Header>,
```
ikopylov: If I understand correctly, this is an in-memory index. BTree is good for disk indexing. For memory there are better data structures: hash map, binary tree.

src/blob.rs
```diff
@@ -1,86 +1,258 @@
...
+            name,
+            index: Default::default(),
+            current_offset: len,
+        })
+    }
+
+    pub fn write(&mut self, record: &mut Record) -> Result<WriteFuture> {
+        record.set_blob_offset(self.current_offset);
+        self.index
+            .insert(record.key().to_vec(), record.header().clone());
```
ikopylov: What happen if the data with the same key already exists in this BLOB?

src/blob.rs
```diff
@@ -1,86 +1,258 @@
...
+        Ok(Self {
+            header: Header::new(),
+            file,
+            name,
+            index: Default::default(),
+            current_offset: len,
+        })
+    }
+
+    pub fn write(&mut self, record: &mut Record) -> Result<WriteFuture> {
```
ikopylov: No synchronization or concurrency support. I guess it will be implemented in the further PRs. Am I right?

src/blob.rs
ikopylov: One more question: what ordering function is used for `Vec<u8>`? I don't see any implementation here.

src/blob.rs
```diff
@@ -1,86 +1,258 @@
...
+            key: key.as_ref().to_vec(),
+            loc: self.locate(&key)?,
+        })
+    }
+
+    fn locate<K>(&self, key: &K) -> Result<Location>
+    where
+        K: AsRef<[u8]> + Ord,
+    {
+        let header = self.index.get(key.as_ref()).ok_or(Error::NotFound)?;
```
ikopylov: You don't read `index` when you open an existing file. This will result in `Error::NotFound`, even when the record is inside the BLOB.

src/blob.rs
```diff
@@ -1,86 +1,258 @@
...
-    /// Reads record data, yields `Ok(Vec<u8>)` if read successful,
-    /// otherwise - `Err`
-    // @TODO more useful result
-    pub fn read(&self, _key: &K) -> Result<Record<K>, ()> {
-        // @TODO implement
-        Ok(Record::new())
+    /// Creates new blob file
+    pub fn open_new(name: FileName) -> Result<Self> {
+        Ok(Self {
+            header: Header::new(),
```
agend: Put high level logic here with stub function:
1) open file
2) write header
3) create index (with result handling)

src/blob.rs
```diff
@@ -1,86 +1,258 @@
...
+            current_offset: 0,
+        })
+    }
+
+    pub fn boxed(self) -> Box<Self> {
+        Box::new(self)
+    }
+
+    pub fn from_file(path: PathBuf) -> Result<Self> {
+        let file = fs::OpenOptions::new()
```
agend: High level logic with stubs functions:
1) open file for blob
2) check header and correctness
3) optional data consistency check for all data
4) open index
5) if open index fails - create a new one from the blob (optional)

src/blob.rs
agend: Yes.

src/blob.rs
```diff
@@ -1,86 +1,258 @@
...
+            header: Header::new(),
+            file,
+            name,
+            index: Default::default(),
+            current_offset: len,
+        })
+    }
+
+    pub fn write(&mut self, record: &mut Record) -> Result<WriteFuture> {
+        record.set_blob_offset(self.current_offset);
```
agend: add bitflags for overwrite as write parameters - will be used later

src/blob.rs
agend: should fail now

src/blob.rs
agend: rename locate - lookup

src/storage.rs
```diff
@@ -73,42 +82,64 @@ where
...
+        self.active_blob
+            .as_mut()
+            .unwrap()
+            .write(record)
+            .map_err(Error::BlobError)
+    }
+
+    fn next_blob_name(&self) -> Result<blob::FileName> {
+        let active_blob_id = self.active_blob.as_ref().map(|blob| blob.id());
+        let blobs_max_id = self.blobs.iter().max_by_key(|blob| blob.id()).map(Blob::id);
```
ikopylov: Suboptimal approach. It is better to keep BLOBs in order and simply add 1 to `active_blob_id` to calculate `next_id`

src/storage.rs
```diff
@@ -155,67 +187,85 @@ where
...
+                        Some(blob)
+                    }
                 } else {
-                    debug!("skipping file \"{}\"", entry.path().display());
+                    max_blob_file_index = Some(id);
+                    active_blob = Some(blob);
                     None
                 }
             })
             .collect();
```
ikopylov: It is better to store BLOBs in order

src/blob.rs
```diff
@@ -1,86 +1,258 @@
...
+            .open(&path)
+            .map_err(Error::FromFile)?;
+        let name = FileName::from_path(&path)?;
+        let len = file.metadata().map_err(Error::FromFile)?.len();
+        // @TODO Scan existing file to create index
+        Ok(Self {
+            header: Header::new(),
+            file,
+            name,
+            index: Default::default(),
```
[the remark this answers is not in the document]
pyakushin: Will be implemented in next issue

src/blob.rs
pyakushin: Implementation of initialization from file will be in the next issue

src/blob.rs
pyakushin: Yes, you are

src/blob.rs
pyakushin: [```Vec::ord(_)```](https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-Ord)

src/storage.rs
pyakushin: Done

src/blob.rs
```diff
@@ -1,86 +1,342 @@
...
 
-/// # Description
-/// # Examples
 #[derive(Debug, Default)]
-struct Header {
-    magic_byte: u64,
-    version: u32,
-    key_size: u32,
-    flags: u64,
+struct Index {
```
ikopylov: 👍

src/blob.rs
```diff
@@ -1,86 +1,342 @@
...
+    headers: BTreeMap<Vec<u8>, record::Header>,
 }
```
ikopylov: I suggest using the range index additionally. It would allow to reject not presented entries faster.

agend: @ikopylov add issue for future plans

src/blob.rs
pyakushin: Why are they better?
[HashMap vs BTreeMap](https://www.reddit.com/r/rust/comments/7rgowj/hashmap_vs_btreemap/)
Rust collections [usecases](https://doc.rust-lang.org/std/collections/index.html)

src/blob.rs
pyakushin: Return ```Error::AlreadyContainsSameKey```

src/blob.rs
agend: this is temporary - don't pick on this - we will implement in future something suitable

src/blob.rs
pyakushin: Read index from existing file is the next issue
Rename done

src/blob.rs
```diff
@@ -1,86 +1,258 @@
...
+    pub fn from_path(path: &Path) -> Result<Self> {
+        let file_name = path
+            .file_name()
+            .ok_or_else(|| Error::PathWithoutFileName(path.to_owned()))?;
+        let fname_str = file_name
+            .to_str()
+            .ok_or_else(|| Error::NonUnicode(file_name.to_owned()))?;
+        let mut parts = fname_str.split('.');
+        let name_prefix = parts.next().unwrap().to_owned();
+        let id = parts.next().unwrap().parse().unwrap();
```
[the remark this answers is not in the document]
pyakushin: reworked

src/blob.rs
pyakushin: done

src/blob.rs
pyakushin: done

src/storage.rs
pyakushin: reworked

src/storage.rs
pyakushin: done

PR #12, opened 2019-04-08
Feature/multithread

pyakushin:


src/blob.rs
```diff
@@ -23,12 +20,30 @@ pub struct Blob {
...
+    current_offset: Arc<Mutex<u64>>,
+}
+
+impl Clone for Blob {
+    fn clone(&self) -> Self {
+        Self {
+            header: self.header.clone(),
+            current_offset: self.current_offset.clone(),
+            name: self.name.clone(),
+            file: self.file.try_clone().unwrap(),
```
agend: sys call detected

pyakushin: Outdated, `Blob` is no longer cloned

src/storage.rs
```diff
@@ -128,21 +232,20 @@ impl Storage {
...
-        K: AsRef<[u8]> + Ord,
-    {
-        if let Some(fut) = self
-            .active_blob
-            .as_ref()
-            .and_then(|active_blob| active_blob.read(&key).ok())
-        {
-            Ok(fut)
+    pub async fn read(&self, key: Vec<u8>) -> Result<Record> {
+        let inner = await!(self.inner.lock());
```
agend: no concurrent reads (

src/blob.rs
```diff
@@ -169,62 +134,57 @@ impl Blob {
...
-        self.current_offset += buf.len() as u64;
-        Ok(WriteFuture {
-            file: self.file.try_clone().map_err(Error::CloneFd)?,
-            buf: Some(buf),
-            current_offset: self.current_offset,
-        })
+        let mut offset = await!(self.current_offset.lock());
+        let bytes_written = self
+            .file
+            .write_at(&buf, *offset)
```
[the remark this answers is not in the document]
pyakushin: Done

src/storage.rs
pyakushin: Agree, RwLock from std needs some adaptation for async/await

src/storage.rs
```diff
@@ -29,11 +41,83 @@ type Result<T> = std::result::Result<T, Error>;
...
-    next_blob_id: usize,
+}
+
+#[derive(Debug)]
+struct Shared {
+    config: Config,
+    next_blob_id: AtomicUsize,
+}
+
+async fn launch_observer<S>(mut spawner: S, storage: Storage)
```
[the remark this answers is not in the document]
pyakushin: Done

src/storage.rs
```diff
@@ -29,11 +41,83 @@ type Result<T> = std::result::Result<T, Error>;
...
+    next_update: Instant,
+}
+
+impl Stream for Observer {
+    type Item = FutureObj<'static, ()>;
+
+    fn poll_next(mut self: Pin<&mut Self>, waker: &Waker) -> Poll<Option<Self::Item>> {
+        println!("Observer poll next");
+        if self.next_update < Instant::now() {
+            self.as_mut().next_update = Instant::now() + Duration::from_millis(1);
```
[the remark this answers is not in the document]
pyakushin: Added to config.
Unfortunately, ```tokio::Interval``` works only on 0.1 Exectors

src/storage.rs
```diff
@@ -29,11 +41,83 @@ type Result<T> = std::result::Result<T, Error>;
...
+            Poll::Pending
+        }
+    }
+}
+
+async fn update_active_blob(s: Arc<Storage>) {
+    println!("Storage update");
+    // @TODO process unwrap explicitly
+    let mut inner = await!(s.inner.lock());
+    let is_full = inner.active_blob.as_ref().unwrap().file_size().unwrap()
```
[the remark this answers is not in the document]
pyakushin: Fixed in this and other places

src/blob.rs
```diff
@@ -169,62 +134,57 @@ impl Blob {
...
-            file: self.file.try_clone().map_err(Error::CloneFd)?,
-            key: key.as_ref().to_vec(),
-            loc: self.lookup(&key)?,
-        })
+    pub async fn read(&self, key: Vec<u8>) -> Result<Record> {
+        let loc = self.lookup(&key)?;
+        let mut buf = vec![0u8; loc.size as usize];
+        let bytes_read = self
+            .file
+            .read_at(&mut buf, loc.offset)
```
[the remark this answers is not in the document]
pyakushin: Done

src/storage/core.rs
```diff
@@ -0,0 +1,471 @@
...
+    shared: Arc<Shared>,
+    inner: Arc<Mutex<Inner>>,
+}
+
+impl Drop for Storage {
+    fn drop(&mut self) {
+        let twins = self.twins_count.fetch_sub(1, Ordering::Relaxed);
+        // 1 is because twin#0 - in observer tread, twin#1 - self
+        if twins <= 1 {
+            self.shared.observer_state.store(false, Ordering::Relaxed);
```
agend: Can we give more straight forward name for `observer_state` something like `need_exit`

src/storage/core.rs
```diff
@@ -0,0 +1,471 @@
...
+    fn new() -> Self {
+        Self {
+            active_blob: None,
+            blobs: Vec::new(),
+            lock_file: None,
+        }
+    }
+
+    fn max_id(&self) -> Option<usize> {
+        let active_blob_id = self.active_blob.as_ref().map(|blob| blob.id());
```
agend: THis is a panic if active blob_id doen't have max id.

src/storage/core.rs
```diff
@@ -0,0 +1,471 @@
...
+        let blobs_max_id = self.blobs.iter().max_by_key(|blob| blob.id()).map(Blob::id);
```
agend: As we store blobs as sorted array - this could be simplified as blobs[blobs.len()-1].id

src/storage/core.rs
```diff
@@ -0,0 +1,471 @@
...
+    let config_max = s.shared.config.max_blob_size.ok_or(Error::Unitialized)?;
+    let is_full = active_size > config_max
+        || active_blob.records_count() as u64
+            >= s.shared.config.max_data_in_blob.ok_or(Error::Unitialized)?;
+
+    if is_full {
+        trace!("is full, replace");
+        let next = s.next_blob_name()?;
+        trace!("next name: {:?}", next);
+        let new_active = Blob::open_new(next)?.boxed();
```
agend: This operation could take a while.  Could we do this outside of lock? We can put check in one lock and if full - prepare new blob and hold lock for the second time and replace. this would be much better it terms of time we hold lock

src/storage/core.rs
pyakushin: No, it returns `Option<usize>`. In case if active blob doesn't have max id, it would be `None`

src/storage/core.rs
pyakushin: Done

src/storage/core.rs
pyakushin: Done

agend: Is this possible at all?

PR #15, opened 2019-04-19
Feature/benchmark

pyakushin:
Completed write part of benchmark. For read and read/write part index must be implemented.

### DIFF
--- a/src/benchmark/writer.rs
+++ b/src/benchmark/writer.rs
@@ -27,4 +27,8 @@ impl Writer {
         println!("write: key {}, value: {}", key.len(), value.len());
         Report::new()
     }
+
+    pub fn close(&mut self) {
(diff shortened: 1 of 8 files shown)


src/benchmark/writer.rs
```diff
@@ -1,39 +1,37 @@
...
-    speed: usize,
     storage: Storage,
 }
 
 impl Writer {
-    pub fn new(speed: usize, tmp_dir: PathBuf) -> Self {
+    pub fn new(tmp_dir: PathBuf) -> Self {
         let storage = Builder::new()
             .blob_file_name_prefix("benchmark")
             .max_blob_size(100_000_000)
```
agend: `max_blob_size`  should be configurable. as well as `max_data_in_blob`

PR #16, opened 2019-04-24
Feature/index

pyakushin:


src/storage/builder.rs
```diff
@@ -92,26 +92,21 @@ impl Builder {
...
-                "blob file format: {}.{{}}.blob",
-                self.config.blob_file_name_prefix.as_ref().unwrap()
-            );
         }
         self
     }
+
+    /// # Description
+    /// Sets key size limit
+    pub fn key_size(mut self, key_size: u64) -> Self {
```
agend: `u64` is to huge for key size) `u8` is too small and `u16` seams to be fine)

src/storage/core.rs
```diff
@@ -137,11 +138,21 @@ impl Storage {
...
-    pub async fn write(self, record: Record) -> Result<()> {
+    /// ```no-run
+    /// block_on(async {
+    ///     let key = 42u64.to_be_bytes().to_vec();
+    ///     let data = b"async written to blob".to_vec();
+    ///     await!(storage.write(key, data))
+    /// )};
+    /// ```
+    pub async fn write(self, key: Vec<u8>, data: Vec<u8>) -> Result<()> {
+        if key.len() as u64 != self.shared.config.key_size.ok_or(Error::Unitialized)? {
```
agend: Опечатка `Unitialized`?

src/storage/core.rs
```diff
@@ -137,11 +138,21 @@ impl Storage {
...
+            return Err(Error::KeySizeMismatch(key.len()));
```
agend: Checking for key size can be put in function. May be it make sense to create some trait for key and not to pass `Vec` every where and we can implement for some basic types (`u64` for example)

src/storage/builder.rs
pyakushin: Agree

src/storage/core.rs
pyakushin: Fixed

PR #17, opened 2019-05-23
Feature/binary search in index file

pyakushin:


src/blob/simple_index.rs
```diff
@@ -0,0 +1,248 @@
...
+            name,
+        })
+    }
+
+    async fn binary_search(mut file: File, key: Vec<u8>) -> Result<RecordHeader> {
+        let header: Header = await!(Self::read_index_header(&mut file))?;
+
+        let mut size = header.records_count;
+        if size == 0 {
+            unimplemented!();
```
agend: why?

src/blob/simple_index.rs
```diff
@@ -0,0 +1,248 @@
...
+
+        let mut start = 0;
+
+        while size > 1 {
+            let half = size / 2;
+            let mid = start + half;
+            let mid_record_header = await!(Self::read_at(&mut file, mid, header))?;
+            let cmp = mid_record_header.key().cmp(&key);
+            debug!("mid read: {:?}", mid_record_header.key());
+            start = if cmp == Greater {
```
agend: May be match is better?

src/blob/simple_index.rs
```diff
@@ -0,0 +1,248 @@
...
+                    .map_err(Error::IO);
+                bunch.sort_by_key(|h| h.key().to_vec());
+                let mut record_header_size = 0;
+                let buf = bunch
+                    .iter()
+                    .filter_map(|h| {
+                        debug!("write ke: {:?}", h.key());
+                        serialize(&h).ok()
+                    })
+                    .fold(Vec::new(), |mut acc, h_buf| {
```
agend: We can pre allocate in advance

src/blob/simple_index.rs
```diff
@@ -0,0 +1,248 @@
...
+    fn dump(&mut self) -> Dump {
+        match &mut self.inner {
+            State::InMemory(bunch) => {
+                let fd_res = fs::OpenOptions::new()
+                    .create(true)
+                    .read(true)
+                    .write(true)
+                    .open(self.name.as_path())
+                    .map_err(Error::IO);
+                bunch.sort_by_key(|h| h.key().to_vec());
```
agend: better to have function here to get buffer

src/blob/simple_index.rs
```diff
@@ -0,0 +1,248 @@
...
+                    Err(e) => return Dump(future::err(e).boxed()),
+                };
+                let inner = State::OnDisk(file.clone());
+                let header = Header {
+                    record_header_size,
+                    records_count: bunch.len(),
+                };
+                let raw_header_res = serialize(&header);
+                self.inner = inner;
+                match raw_header_res {
```
agend: function here as well for dumping ready to use buffers

src/blob/simple_index.rs
```diff
@@ -89,7 +89,7 @@ impl SimpleIndex {
 
         let mut size = header.records_count;
         if size == 0 {
-            unimplemented!();
+            error!("provided empty key");
```
agend: empty key was provided

PR #31, opened 2019-09-24
read/read_with/read_all

pyakushin:


src/blob/entry.rs
```diff
@@ -0,0 +1,210 @@
+use super::prelude::*;
+use crate::prelude::*;
+
+use std::slice::Iter;
+
+/// [`Entry`] is a [`Future`], which contains header and metadata of the record,
+/// but does not contain all of the data in memory. To resolve [`Entry`] to data,
+/// you should poll it.
+/// [`Entry`]: struct.Entry.html
+pub struct Entry {
```
agend: Consider to rename

src/storage/read_all.rs
```diff
@@ -0,0 +1,114 @@
+use super::prelude::*;
+
+/// @TODO
+#[derive(Debug)]
+pub struct ReadAll<'a, K> {
+    key: &'a [u8],
+    inner: &'a Storage<K>,
+    state: RefCell<State<'a>>,
+    ready_entries: Vec<Entry>,
```
agend: Looks like `VecDeque` is better here

PR #33, opened 2019-11-14
skip empty blobs while init on existing work dir

pyakushin:


src/blob/core.rs
```diff
@@ -107,6 +107,11 @@ impl Blob {
         let file: File = File::from_std_file(Self::open_file(&path)?)?;
         let name = FileName::from_path(&path)?;
         let len = file.metadata()?.len();
+        let header = Header::new();
+        if len <= bincode::serialized_size(&header)? {
+            fs::remove_file(&path)?;
```
agend: Why do you remove original file when it's empty file. Better just to skip it - not destructive behaviour

pyakushin: I read blob file and if it empty, just skip index generation. So if empty blob file is the last, it will be used as active blob.

src/error.rs
```diff
@@ -120,6 +120,8 @@ pub enum Kind {
     WrongFileNamePattern(PathBuf),
     /// Conversion error
     Conversion(String),
+    /// Closed blob has no records
+    EmptyBlob,
```
agend: You not using this type now.

PR #38, opened 2019-12-11
Make check existence optional

pyakushin:


src/storage/core.rs
```diff
@@ -384,6 +385,10 @@ impl<K> Storage<K> {
         debug!("async init blobs from file");
         futures.try_collect().await.map_err(Error::new)
     }
+
+    fn existence_check_is_on(&self) -> bool {
```
agend: `allow_duplicates` is good name why do you want it to chage?

PR #45, opened 2020-03-10
Storage fast check contains

pyakushin:


src/storage/core.rs
```diff
@@ -412,17 +412,38 @@ impl<K> Storage<K> {
...
+        in_active
+    }
+
+    /// `fast_check` is used to check whether a key is in storage.
     /// Uses bloom filter under the hood, so false positive results are possible,
     /// but false negatives are not.
-    /// In other words, `contains` returns either "possibly in storage" or "definitely not".
-    pub async fn contains(&self, key: impl Key) -> bool {
+    /// In other words, `fast_check` returns either "possibly in storage" or "definitely not".
+    pub async fn fast_check(&self, key: impl Key) -> bool {
```
agend: if this is bloom filter - lets call this method accordingly (`check_bloom`). In pearl as well please)

src/blob/core.rs
```diff
@@ -191,8 +191,8 @@ impl Blob {
         self.index.get_entry(key, self.file.clone())
     }
 
-    async fn lookup(&self, key: &[u8], meta: Option<&Meta>) -> Option<Location> {
-        if self.contains(key) {
+    async fn get_location(&self, key: &[u8], meta: Option<&Meta>) -> Option<Location> {
```
agend: why did you change from lookup to get_location?

pyakushin: No particular reason, thought it would be more clear

PR #48, opened 2020-05-19
File data sync

pyakushin:


src/blob/file.rs
```diff
@@ -4,130 +4,59 @@ const WOULDBLOCK_RETRY_INTERVAL_MS: u64 = 10;
 
 #[derive(Debug, Clone)]
 pub(crate) struct File {
-    pub(crate) read_fd: Arc<fs::File>,
-    pub(crate) write_fd: Arc<Mutex<fs::File>>,
+    read_fd: Arc<StdFile>, // requires only for read_at/write_at methods
```
agend, line 7: `read_fd` makes impression that it used for read-only operations and `write_fd` for write-only. May be better to name it something like this: `no_lock_fd` and `write_fd` to something like `cursor_fd`

src/blob/index/simple.rs
```diff
@@ -128,7 +123,7 @@ impl Simple {
         ))
     }
 
-    pub async fn load_records(mut file: &File) -> Result<Vec<RecordHeader>> {
+    pub async fn load_records(file: &File) -> Result<Vec<RecordHeader>> {
         trace!("seek to file start");
         file.seek(SeekFrom::Start(0)).await?;
```
agend, line 128: Is there a race condition with line 136? As after you seek to the position someone can change location before you actually read it. Or it is done during init state?

agend: same question to each place with `seek` operations

src/blob/file.rs
pyakushin: fixed in #49 PR

src/blob/index/simple.rs
pyakushin: seek operations removed in #49

PR #49, opened 2020-05-25
make bloom filter optional

pyakushin:


src/blob/index/simple.rs
```diff
@@ -39,22 +40,32 @@ pub(crate) enum State {
...
 
 impl Simple {
-    pub(crate) fn new(filter_config: &Config, name: FileName) -> Self {
-        let filter = Bloom::new(filter_config);
+    pub(crate) fn new(filter_config: Option<Config>, name: FileName) -> Self {
+        let filter_is_on = filter_config.is_some();
+        let filter = if let Some(config) = filter_config {
+            Bloom::new(config)
+        } else {
+            Bloom::default()
```
agend, line 48: add comments how it will works for different cases.

src/storage/core.rs
```diff
@@ -433,21 +436,24 @@ impl<K> Storage<K> {
...
     /// In other words, `check_bloom` returns either "possibly in storage" or "definitely not".
-    pub async fn check_bloom(&self, key: impl Key) -> bool {
+    pub async fn check_bloom(&self, key: impl Key) -> Option<bool> {
         trace!("[{:?}] check in blobs bloom filter", &key.to_vec());
         let inner = self.inner.safe.lock().await;
         let in_active = inner
             .active_blob
             .as_ref()
-            .map_or(false, |active_blob| active_blob.check_bloom(key.as_ref()));
+            .map_or(Some(false), |active_blob| {
```
agend, line 449: is there on pearl level options to skip bloom index lookup?

pyakushin: Yes, if filter is opt out, `check_bloom` will return `None`.

PR #52, opened 2020-08-05
track file len in memory

pyakushin:


src/blob/file.rs
```diff
@@ -30,38 +31,27 @@ impl File {
         Self::from_tokio_file(file, ioring).await
     }
 
-    pub(crate) async fn metadata(&self) -> IOResult<Metadata> {
-        self.write_fd.read().await.metadata().await
+    pub fn len(&self) -> u64 {
```
agend, line 34: I think size more appropriate name

src/blob/file.rs
```diff
@@ -30,38 +31,27 @@ impl File {
...
-    // }
+    pub(crate) async fn _metadata(&self) -> IOResult<Metadata> {
+        self.write_fd.read().await.metadata().await
+    }
 
     pub(crate) async fn write_at(&self, buf: &[u8], offset: u64) -> IOResult<usize> {
         let compl = self.ioring.write_at(&*self.no_lock_fd, &buf, offset);
-        compl.await
+        let add_len = compl.await?;
+        self.len.fetch_add(add_len as u64, ORD);
```
agend, line 45: Here will be a miscalculation when you will write in the middle of file

src/blob/file.rs
```diff
@@ -76,92 +66,37 @@ impl File {
...
-//                 warn_and_wake(cx.waker().clone());
-//                 Poll::Pending
-//             }
-//             Err(e) => Poll::Ready(Err(e)),
-//             Ok(n) => Poll::Ready(Ok(buf[0..n].to_vec())),
-//         }
-//     }
-// }
+    fn len_from_u64(len: u64) -> Arc<AtomicU64> {
+        Arc::new(AtomicU64::new(len))
```
agend, line 100: one liner... should we use separate function for it?

PR #56, opened 2020-08-17
Rework read all

pyakushin:


src/blob/core.rs
```diff
@@ -183,31 +183,54 @@ impl Blob {
...
     }
 
     #[inline]
-    pub(crate) fn read_all<'a>(&'a self, key: &'a [u8]) -> Entries<'a> {
-        self.index.get_entries(key, self.file.clone())
+    pub(crate) async fn read_all(&self, key: &[u8]) -> Result<Option<Vec<Entry>>> {
+        let headers = self.index.get_all(key).await?;
+        Ok(headers.map(|h| {
+            debug!("blob core read all {} headers", h.len());
+            Self::headers_to_entries_with_meta(h, self.file.clone())
```
agend, line 205: I think using method here is redundant

src/blob/core.rs
```diff
@@ -183,31 +183,54 @@ impl Blob {
...
+            .unwrap()
             .into_data();
         debug!("blob read any entry loaded bytes: {}", buf.len());
         Ok(buf)
     }
 
     #[inline]
-    pub(crate) fn read_all<'a>(&'a self, key: &'a [u8]) -> Entries<'a> {
-        self.index.get_entries(key, self.file.clone())
+    pub(crate) async fn read_all(&self, key: &[u8]) -> Result<Option<Vec<Entry>>> {
```
agend, line 201: should we rename it to `get_all_entries`

src/blob/core.rs
```diff
@@ -183,31 +183,54 @@ impl Blob {
...
+    }
+
+    fn headers_to_entries_with_meta(headers: Vec<RecordHeader>, file: File) -> Vec<Entry> {
+        headers
+            .into_iter()
+            .map(|header| Entry::new(header, file.clone()))
+            .collect()
     }
 
     async fn get_any_entry(&self, key: &[u8], meta: Option<&Meta>) -> Result<Option<Entry>> {
```
agend, line 216: it looks like to get particular entry with matching meta but not an any. Should we rename it to `get_entry`

src/blob/core.rs
```diff
@@ -234,8 +257,10 @@ impl Blob {
 
     async fn find_entry<'a>(ents: Entries<'a>, meta: &'a Meta) -> Result<Option<Entry>> {
```
agend, line 258: Seams this function is not used any more

src/blob/index/simple.rs
```diff
@@ -462,7 +502,7 @@ impl Index for Simple {
 
     async fn count(&self) -> Result<usize> {
         match &self.inner {
-            State::InMemory(headers) => Ok(headers.len()),
+            State::InMemory(headers) => Ok(headers.values().fold(0, |acc, x| acc + x.len())),
```
agend, line 505: I think `IndexHeader` should have records count.

src/blob/core.rs
pyakushin: I use it in other place [here](https://github.com/qoollo/pearl/pull/56/files#diff-f2bad8d74f50db884e82f478e0f8ceebR242)
Nested closures are hard to read (:

src/blob/core.rs
pyakushin: I don't know, we don't have other plans for `read_all` fn name

src/blob/core.rs
pyakushin: Agree

src/blob/core.rs
pyakushin: Removed

src/blob/index/simple.rs
pyakushin: Done

PR #64, opened 2020-11-09
Feature no aio

pyakushin:


src/blob/core.rs
```diff
@@ -46,21 +47,51 @@ impl Blob {
...
 
+    #[cfg(feature = "aio")]
     async fn write_header(&mut self) -> Result<()> {
         let buf = serialize(&self.header)?;
         let offset = self.file.write_append(&buf).await? as u64;
         self.update_offset(offset).await;
         Ok(())
     }
 
+    #[cfg(not(feature = "aio"))]
```
agend, line 82: exactly the same code - look not good. Is there a way to fix it?

PR #66, opened 2020-11-17
Make usage of the AIO optional

pyakushin:


src/blob/file.rs
```diff
@@ -112,7 +157,17 @@ impl File {
     }
 
     pub(crate) async fn fsyncdata(&self) -> IOResult<()> {
-        let compl = self.ioring.fdatasync(&*self.no_lock_fd);
-        compl.await
+        if let Some(ref ioring) = self.ioring {
+            let compl = ioring.fdatasync(&*self.no_lock_fd);
+            compl.await
+        } else {
+            let (tx, rx) = channel();
```
agend, line 164: I think could be as well as designated func

tests/common.rs
```diff
@@ -71,15 +71,16 @@ pub async fn create_test_storage(
...
         .work_dir(&path)
         .blob_file_name_prefix("test")
         .max_blob_size(max_blob_size)
         .max_data_in_blob(100_000)
         .set_filter_config(Default::default())
-        .allow_duplicates();
-    let ioring = rio::new().expect("create uring");
-    let mut storage = builder.build(ioring).unwrap();
+        .allow_duplicates()
+        .enable_aio(ioring);
```
agend, line 82: Could also provide func to test for support of ioring.

pyakushin: ioring is created outside of pearl, so if user can't create rio, he won't be able to enable aio
```
let rio: Result<Rio, Error> = rio::new();
let storage = if let Ok(ioring) = rio {
    builder
        .enable_aio(ioring)
        .build()
} else {
    builder.build()
};
```

src/blob/file.rs
pyakushin: Done

PR #73, opened 2020-12-21
validate-index-file

idruzhitskiy:


src/blob/index/simple.rs
```diff
@@ -138,14 +152,19 @@ impl Simple {
             .await
             .with_context(|| format!("file open failed {:?}", self.name.to_path()))?;
         let size = file.write_append(&buf).await?;
+        file.write_append(&MAGIC_NUMBER.to_be_bytes()).await?;
```
ikopylov, line 155: I suggest to write it to the header, not to the end of the file. With the current approach, it will be potentially more difficult to extend the index format

src/blob/index/simple.rs
```diff
@@ -10,11 +12,12 @@ pub(crate) struct Simple {
     ioring: Option<Rio>,
 }
 
-#[derive(Debug, Deserialize, Default, Serialize, Clone)]
+#[derive(Debug, Deserialize, Serialize, Clone)]
 pub(crate) struct IndexHeader {
     pub records_count: usize,
     pub record_header_size: usize,
     pub filter_buf_size: usize,
+    pub hash: Vec<u8>,
```
ikopylov, line 20: Index header has changed, so old index files are no longer valid. This is problem. In the current state we can ignore it, but I suggest to add a version field at the very beggining of the index header. This will allow to distinguish between versions in the future

ikopylov: @pyakushin what do you think?

src/blob/index/simple.rs
```diff
@@ -160,6 +186,25 @@ impl Simple {
         self.filter = Bloom::from_raw(buf_ref)?;
         Ok(())
     }
+
+    async fn read_header_from_file(file: File) -> Result<Vec<u8>> {
+        let mut buf = file.read_all().await?;
+        let end_magic: Vec<_> = buf.drain(buf.len() - std::mem::size_of::<u64>()..).collect();
```
pyakushin, line 192: I think it's better to use split. And don't create new `Vec<_>` but just compare slices below

pyakushin (approved):

PR #80, opened 2021-01-29
New error for unavailable disk

### LINKED ISSUE #77: Return a special error code from the write method if the physical disk is unavailable
Bob needs this to handle a disk failure

Justarone:
Seems like it will close #77


src/error.rs
```diff
@@ -81,8 +85,8 @@ pub enum Kind {
     WorkDirInUse,
     /// Happens when try to write/read from work dir that doesn't exist.
     /// In case when work dir wasn't created or disk was unmounted.
-    /// Contains path to failed work dir.
-    WorkDirUnavailable(String),
+    /// Contains path to failed work dir, IOError description and IOErrorKind.
+    WorkDirUnavailable((String, String, IOErrorKind)),
```
pyakushin, line 89: Should be ```WorkDirUnavailable(String, String, IOErrorKind)```

pyakushin: And I think it's convenient to make this variant a struct:
```
enum Kind {
    WorkDirUnavailable{ path: PathBuf, msg: String, io_err_kind: IOErrorKind }
}
```

src/storage/core.rs
```diff
@@ -203,7 +203,22 @@ impl<K> Storage<K> {
             .active_blob
             .as_mut()
             .ok_or_else(Error::active_blob_not_set)?;
-        blob.write(record).await
+        match blob.write(record).await {
```
pyakushin, line 206: It is wrong to downcast inside pearl. We can return the error we want from blob.write().

Justarone: hm, shouldn't this error then be in bob? or we should define this error in pearl but use in bob?

Justarone: And there is one more question: if we don't downcast this error in pearl, then this error will be received in disk_controller (according to todays version of bob) and in there we can organize processing of os error without converting to WorkDirUnavailable (Maybe I misunderstood you, waiting for an answer)

Justarone:
I want to notice one fact:
If you're trying to read unplugged disk data and it is NOT cached in RAM, you'll get OsError(5) ("Input/Output Error")
I want to discuss: should we try to detect disk unplug event during read operation or it's enough to do that during write?

src/storage/core.rs
```diff
@@ -203,7 +203,25 @@ impl<K> Storage<K> {
...
+            Ok(_) => Ok(()),
+            Err(err) => match err.downcast::<Error>() {
```
idruzhitskiy, line 208: I think you can reduce nesting here, with something like that: 
```
     blob.write(record).await.or_else(|e| {
            let err = e.downcast::<Error>()?;

            if let ErrorKind::FileUnavailable(kind) = err.kind() {
                let dir = self.inner.config.work_dir().ok_or(Error::uninitialized())?;
                return Err(Error::work_dir_unavailable(dir, err.to_string(), *kind).into());
            }

            Err(err.into())
        })

```
Not sure about using "if let return" or "match", though

idruzhitskiy: Probably it's even better to completely eliminate multiple return points by making err mutable and assigning it in the branch

src/blob/core.rs
```diff
@@ -180,7 +180,18 @@ impl Blob {
...
         let buf = record.to_raw()?;
-        let bytes_written = self.file.write_append(&buf).await? as u64;
+        let bytes_written = self
+            .file
+            .write_append(&buf)
+            .await
+            .map_err(|e| -> anyhow::Error {
+                match e.kind() {
+                    kind if kind == IOErrorKind::Other || kind == IOErrorKind::NotFound => {
+                        Error::file_unavailable(kind).into()
```
pyakushin, line 190: `?` performs `into()` conversion under the hood, are you sure it is necessary here?

src/blob/file.rs
```diff
@@ -47,14 +47,26 @@ impl File {
...
+    fn update_size(&self, diff: u64) {
+        self.size.fetch_add(diff, Ordering::SeqCst);
+    }
+
     pub(crate) async fn write_append_sync(&self, buf: &[u8]) -> IOResult<usize> {
-        let offset = self.size.fetch_add(buf.len() as u64, Ordering::SeqCst);
-        self.write_at_sync(offset, buf).await
+        let offset = self.size.load(Ordering::SeqCst);
+        self.write_at_sync(offset, buf).await.map(|bytes_written| {
+            self.update_size(bytes_written as u64);
```
pyakushin, line 57: What if other task will change size after `load` and before `update_size`?

pyakushin (changes requested):

src/blob/core.rs
Justarone: I convert some of OS errors which may occur due to disk crash to Pearl error which indicates that. There are different errors in branch so I need to convert error to `anyhow::Error` inside of closure, and that's before `?` operator action.

pyakushin (approved):

PR #86, opened 2021-03-21
Add an option to ignore corrupted blobs

Justarone:
This will be helpful, if some other application wants to restart storage on the same disk after its disconnection (disk fault). This option let storage to start with normal blobs only instead of blocking if there are some (even 1) corrupted blobs (but probability of corrupted closed files after merging #85 PR will be reduced much)


pyakushin (approved):
lgtm

PR #90, opened 2021-03-28
Add check to avoid unneeded fsync operations

Justarone:
Add check before performing dump steps AND make a bit more informative logs


pyakushin (approved):
LGTM

PR #91, opened 2021-03-28
Change `safe` field of `Inner` struct on `RwLock` (instead of `Mutex`)

Justarone:


src/storage/core.rs
```diff
@@ -463,7 +462,7 @@ impl<K> Storage<K> {
                 Error::from(ErrorKind::Uninitialized)
             })?
             .boxed();
-        let mut safe_locked = self.inner.safe.lock().await;
+        let mut safe_locked = self.inner.safe.write().await;
```
pyakushin, line 465: I suggest to rename variable to smth like just `safe` or `inner`

src/storage/core.rs
```diff
@@ -537,7 +536,7 @@ impl<K> Storage<K> {
     }
 
     async fn contains_with(&self, key: &[u8], meta: Option<&Meta>) -> Result<bool> {
-        let inner = self.inner.safe.lock().await;
+        let inner = self.inner.safe.write().await;
```
pyakushin, line 539: Are you sure `write` is needed here not `read`?

PR #93, opened 2021-03-31
B+ tree file index organization

### LINKED ISSUE #84: Implement B+ Tree index

Justarone:
closes #84


Justarone:
16.03 - 3h (investigating solutions of index disk storing)
23.03 - 1h (b+-tree index research)
24.03 - 5h (indices restructurization)
25.03 - 2h (indices restructurization)
28.03 - 1.5h (finish restructuring and start b+tree)
28.03 - 0.5h (b+-tree indices implementation)
31.03 - 4.5h (b+-tree indices implementation)

src/blob/index/bptree/core.rs
```diff
@@ -0,0 +1,429 @@
...
+        TreeMeta::from_raw(&buf).map_err(Into::into)
+    }
+
+    fn serialize(
+        headers_btree: &InMemoryIndex,
+        filter: &Bloom,
+    ) -> Result<Option<(IndexHeader, TreeMeta, Vec<u8>)>> {
+        debug!("blob index bptree serialize headers");
+        if let Some(record_header) = headers_btree.values().next().and_then(|v| v.first()) {
+            let record_header_size = record_header.serialized_size().try_into()?;
```
idruzhitskiy, line 278: IMO it's better to use builder pattern here, with object holding data for all methods needed. You can even add simple enum of states to ensure proper methods order (like "add_nonleaves" can be used only after "add_leaves" etc)

src/blob/index/header.rs
```diff
@@ -0,0 +1,67 @@
...
+
+#[derive(Debug, Deserialize, Serialize, Clone)]
+pub(crate) struct IndexHeader {
+    pub records_count: usize,
+    // contains serialized size of record headers, which allows to calculate offset in
+    // case of `OnDisk` state of indices
+    pub record_header_size: usize,
+    pub filter_buf_size: usize,
+    pub hash: Vec<u8>,
+    pub(crate) version: u64,
```
agend, line 11: @pyakushin Looks like version as u64 and written as u8 is overkill here ;) Should be it optimized now?

src/blob/index/bptree/node.rs
```diff
@@ -0,0 +1,117 @@
...
+
+impl Node {
+    pub(super) fn new(keys: Vec<Vec<u8>>, offsets: Vec<u64>) -> Self {
+        assert!(keys.len() != 0);
+        assert_eq!(keys.len() + 1, offsets.len());
+        Self { keys, offsets }
+    }
+
+    pub(super) fn key_offset(&self, key: &[u8]) -> u64 {
+        let mut left = 0i32;
```
idruzhitskiy, line 17: You can use "self.data.binary_search_by". It returns either found elem or position at which elem can be inserted (which can be used in case nothing is found to return offset)

src/blob/index/bptree/node.rs
```diff
@@ -0,0 +1,117 @@
...
+        } else {
+            self.offsets[(left + 1) as usize] as u64
+        }
+    }
+
+    pub(super) fn serialize(&self) -> Result<Vec<u8>> {
+        let mut buf = Vec::new();
+        let meta = NodeMeta::new(self.keys.len() as u64);
+        buf.extend_from_slice(&serialize(&meta)?);
+        self.keys.iter().fold(&mut buf, |acc, key| {
```
idruzhitskiy, line 38: Why not flat_map?

src/blob/index/bptree/core.rs
```diff
@@ -0,0 +1,429 @@
+/// structure of b+-tree index file from the beginning:
+/// 1. Header
+/// 2. Bloom filter
+/// 3. Sorted by keys array of record headers (serialized size of one record header is 61 bytes)
+/// 4. Metadata
+/// 5. Sorted by keys array of pairs `key - offset of record header in file` (these are leaf nodes
```
agend, line 6: `key - offset of record header in file` does it mean `key` and `offset of record header in file` ?

agend:
@pyakushin @Justarone It would be nice to put bptree code as separate crate for general use-case and used by pearl. It shouldn't be too much of efforts, right?

src/blob/index/bptree/node.rs
Justarone: I am not sure, that I understood you right, but if I'll iterate through keys with flat_map (and get something like `key_byte` sequence) I'll need to collect there somewhere, because I can't extend from iterator (and this collection step will give one more memory allocation)

src/blob/index/bptree/core.rs
```diff
@@ -0,0 +1,429 @@
+/// structure of b+-tree index file from the beginning:
+/// 1. Header
+/// 2. Bloom filter
```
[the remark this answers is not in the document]
Justarone: I left logic of old `simple` index, but I agree, that maybe we should separate it (and also separate moments, where we send it to file (do it somewhen later). With this we'll filter some requests to not presented keys (=> reduce disk operations)), because it doesn't load memory as mush as btree with indices

src/blob/index/bptree/core.rs
```diff
@@ -0,0 +1,429 @@
...
+    async fn from_records(
+        path: &Path,
+        ioring: Option<Rio>,
+        headers: &InMemoryIndex,
+        filter: &Bloom,
+    ) -> Result<Self> {
+        let res = Self::serialize(headers, filter)?;
+        if res.is_none() {
+            error!("Indices are empty!");
+            return Err(anyhow!("empty in-memory indices".to_string()));
```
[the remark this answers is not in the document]
Justarone: That's part of old logic (from `simple` index). I don't know, why it was implemented that way, but I have a guess: empty BTreeMap doesn't need much memory, but InMemory state allow us to add keys and we don't proceed useless disk operations

src/blob/index/bptree/core.rs
Justarone: Yes, you're right, I have chosen bad separator for pair, I change that

src/blob/index/bptree/core.rs
```diff
@@ -0,0 +1,429 @@
...
+        let (mut header, metadata, buf) = res.expect("None case is checked");
+        let _ = std::fs::remove_file(path);
+        let file = File::create(path, ioring)
+            .await
+            .with_context(|| format!("file open failed {:?}", path))?;
+        file.write_append(&buf).await?;
+        header.written = 1;
+        let serialized_header = serialize(&header)?;
+        file.write_at(0, &serialized_header).await?;
+        file.fsyncdata().await?;
```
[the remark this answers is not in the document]
Justarone: In my opinion we do should fsync when we create file index from records because:
1. If higher-level struct requested creation `from_records` (not `from_file`) then some data changed => then we need to rewrite file (=> then we need to fix state (read (2))
2. Once data is on disk, it's read-only (write operations are not permitted), so if we fsync at the start we don't need to care about when we should do it and potential chance to lose all data  in case of unexpected disk dropping out or something like that (remember about page cache: the data in index file without fsync will likely be corrupted because of dirty pages in RAM, that are not synced with disk)

BUT! I agree that it wouldn't be redundant to add flag in `IndexStruct`: if there were any successful write operations since creating InMemoryIndex - we should create file index from records, otherwise - we can create it from file!
What do you think?

Justarone:
@agend I agree that it would be cool to create separate crate with bptree indices, I am ready to work on that (seems like I should create more general interface and make this struct generic, but that's not a big problem I think)

src/blob/index/bptree/node.rs
idruzhitskiy: You can extend from iterator by `extend` method, why not?

Justarone: Forgot about this method, did it!

src/blob/index/bptree/core.rs
Justarone: I implemented it, but need your review (does it look strange?)

src/blob/index/bptree/core.rs
```diff
@@ -0,0 +1,283 @@
+/// structure of b+-tree index file from the beginning:
+/// 1. Header
+/// 2. Bloom filter
+/// 3. Sorted by keys array of record headers (serialized size of one record header is 61 bytes)
+/// 4. Metadata
+/// 5. Sorted by keys array of pairs `(key, offset of record header in file)` (these are leaf nodes
```
agend, line 6: Are there any consistency control on block level? I saw hash for header.

src/blob/index/bptree/core.rs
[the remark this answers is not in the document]
agend: I need to review index life-cycle one more time. I will come back later this week. @pyakushin Could you check it as well?

src/blob/index/bptree/serializer.rs
```diff
@@ -0,0 +1,269 @@
...
+
+        let hs = header.serialized_size()? as usize;
+        let fsize = header.filter_buf_size;
+        let msize = meta_buf.len();
+        let data_size = hs + fsize + headers_size + msize + leaves_buf.len() + tree_buf.len();
+        let mut buf = Vec::with_capacity(data_size);
+        serialize_into(&mut buf, &header)?;
+        buf.extend_from_slice(&filter_buf);
+        Self::append_headers(self.headers_btree, &mut buf)?;
+        buf.extend_from_slice(&meta_buf);
```
agend, line 71: Does it make sense to extend once for all three appends to avoid possible multiple allocations?

src/blob/index/core.rs
```diff
@@ -0,0 +1,290 @@
...
+                    v.push(h);
+                    trace!("capacity growth: {}", v.capacity() - old_capacity);
+                    mem.records_allocated += v.capacity() - old_capacity;
+                } else {
+                    if mem.records_count == 0 {
+                        // record header contains key as Vec<u8>, h.key().len() - data on the heap
+                        mem.record_header_size = size_of::<RecordHeader>() + h.key().len();
+                        // every entry also includes data on heap: capacity * size_of::<RecordHeader>()
+                        mem.key_size = h.key().len();
+                        mem.btree_entry_size =
```
agend, line 207: this size is calculated more then once with same code. check `compute_mem_attrs`

src/blob/index/bptree/core.rs
Justarone: There is consistency control on the whole file level. I think, that's enough to be sure that every block is consistent (but maybe I misunderstood you)

src/blob/index/bptree/serializer.rs
Justarone: So you suggest to chain filter_buf, meta_buf... with iterator and extend from it  for example? I am sure, that there wouldn't be redundant allocations because I allocate buffer with exact needed size (check line 67)

agend: Missed it - my bad 🤦‍♂️

src/blob/index/bptree/core.rs
```diff
@@ -0,0 +1,283 @@
...
+        {
+            let headers = self.read_headers(fh_offset, amount as usize).await?;
+            Ok(Some(headers))
+        } else {
+            Ok(None)
+        }
+    }
+
+    async fn get_records_headers(&self) -> Result<(InMemoryIndex, usize)> {
+        let mut buf = self.file.read_all().await?;
```
agend, line 101: On start it will create lots of iops to read whole indexes and will result in slow start. Should we consider making it optional?

agend (commented):
> @agend I agree that it would be cool to create separate crate with bptree indices, I am ready to work on that (seems like I should create more general interface and make this struct generic, but that's not a big problem I think)

will you close this one then? and move btree stuff to separate repo?

Justarone:
> > @agend I agree that it would be cool to create separate crate with bptree indices, I am ready to work on that (seems like I should create more general interface and make this struct generic, but that's not a big problem I think)
> 
> will you close this one then? and move btree stuff to separate repo?

I think, I will edit this request (because I also separated index from file index in this request, so I will remove bptree implementation, move it to another crate and in this request will include that crate, I think)

src/blob/index/bptree/serializer.rs
```diff
@@ -0,0 +1,269 @@
...
+}
+
+pub(super) struct Serializer<'a> {
+    headers_btree: &'a InMemoryIndex,
+    header_stage: Option<HeaderStage>,
+    leaves_stage: Option<LeavesStage>,
+    tree_stage: Option<TreeStage>,
+}
+
+impl<'a> Serializer<'a> {
```
idruzhitskiy, line 29: Looks fine, but there is one problem: sequence of stages are checked at runtime, while it's possible to check at compile time. To do that, consider making stages contain needed info (like headers_btree), and make separate implementations, in which there will be only valid order of method invocations. E.g. some stub (optional) will contain method `header_stage`, which will produce instance of `HeaderStage`. And implementation of `HeaderStage` will contain `leaves_stage` method, producing `LeavesStage` and so on. That way you don't need to check, whether previous stages were completed - you know they were.

src/blob/index/bptree/core.rs
Justarone: I don't understand you. How we can make it optional?
Current version of pearl calls this method only once during init: for active blob if there is one. We can make active blob optional for `init_from_existing`, in this case pearl will be read-only, but there won't be any `get_records_headers` operations

Justarone:
@agend I have thought about moving bptree indices to another crate and have one problem: We use our own File implementation and I also use it in indices, so I need to do something with it. I see 3 variants:
1. Duplicate `File` impl in both crates (that's ugly)
2. Move File to separate crate and use it in Pearl and BPTree (But crate with File will be too small)
3. Move File in BPTree crate (but that would be weird import)

I don't know, how to move it to separate crate in appropriate way, so hope for your help

agend:
Reading whole file just for headers could be too much. That is what I’m talking about.
I haven’t checked how it is used bu other parts though.

src/blob/index/bptree/core.rs
Justarone: In older version of indices they consisted of header, filter, and record headers. New version (bptree) consists of header, meta (I changed bloom filter of Vec<u8> and make serialization of filter on higher level, because file index doesn't need to know about it), record headers and b+ tree. New version size ~ 1.2 * size of simple indices version

In older version we read the whole file for hash validation. Here I do the same thing (check validate_header function: it also validates if the hash in header is correct). Do you want to make hash check optional OR to have 2 different hashes (for record headers array and for tree)?

agend: It would be nice to check how it's done by other storage systems. Simple solution - make it optional. But would it be nice to check how it's done by others.

agend:
> * Duplicate `File` impl in both crates (that's ugly)
> * Move File to separate crate and use it in Pearl and BPTree (But crate with File will be too small)
> * Move File in BPTree crate (but that would be weird import)

I agree that seperate crate would be too much. File impl is not small so duplicate looks really ugly. I vote for weird import :) . Any way pearl will tied to bptree crate

src/blob/index/bptree/core.rs
agend: Some FS have CRC on block level (e.g. ZFS, btrfs), but some do not (e.g. ext4)

agend: Postgresql added block level checksum since 9.3 https://postgreshelp.com/postgresql-checksum/

src/blob/index/bptree/core.rs
```diff
@@ -0,0 +1,287 @@
...
+            .collect::<Result<Vec<(Vec<u8>, u64)>>>()
+    }
+
+    fn with_amount(
+        &self,
+        header_pointers: Vec<(Vec<u8>, u64)>,
+        mid: usize,
+        buf_end: u64,
+    ) -> (u64, u64) {
+        let last_rec = mid + 1 >= header_pointers.len() && (buf_end >= self.metadata.tree_offset);
```
pyakushin, line 214: I suggest to rename `last_rec` -> `is_last_rec`

pyakushin (commented):
I think maybe you should introduce struct Leaf {..}

src/blob/index/bptree/node.rs
```diff
@@ -0,0 +1,103 @@
...
+
+#[derive(Debug, PartialEq)]
+pub(super) struct Node {
+    keys: Vec<Vec<u8>>,
+    offsets: Vec<u64>,
+}
+
+impl Node {
+    pub(super) fn new(keys: Vec<Vec<u8>>, offsets: Vec<u64>) -> Self {
+        assert!(keys.len() != 0);
```
pyakushin, line 11: I think it's a not good practice to use assertions in working code

pyakushin: And btw, in that case you should use `!keys.is_empty()` (:

src/blob/index/bptree/node.rs
```diff
@@ -0,0 +1,103 @@
...
+        let keys = keys_buf
+            .chunks(key_size as usize)
+            .map(|key| key.to_vec())
+            .collect();
+        let offsets_buf_size = (meta.size + 1) as usize * std::mem::size_of::<u64>();
+        let (offsets_buf, _rest_buf) = rest_buf.split_at(offsets_buf_size);
+        let offsets = offsets_buf
+            .chunks(std::mem::size_of::<u64>())
+            .map(|bytes| deserialize::<u64>(bytes).map_err(Into::into))
+            .collect::<Result<Vec<u64>>>()?;
```
pyakushin, line 49: Maybe it's more convenient to use here `try_collect()`

src/blob/index/bptree/serializer.rs
```diff
@@ -0,0 +1,275 @@
...
+impl<'a, K: AsRef<[u8]> + 'a, V: Serialize + 'a, It: Iterator<Item = (K, &'a Vec<V>)> + Clone>
+    Serializer<'a, K, V, It>
+{
+    pub(super) fn new(headers_iter: It) -> Self {
+        Self { headers_iter }
+    }
+    pub(super) fn header_stage(self, meta: Vec<u8>) -> Result<HeaderStage<'a, K, V, It>> {
+        if let Some((key, Some(record_header))) = self
+            .headers_iter
+            .clone()
```
pyakushin, line 65: You don't need a `clone()` here, because you can declare `self` as mutable

src/blob/index/bptree/serializer.rs
```diff
@@ -0,0 +1,275 @@
...
+        leaves_buf
+    }
+}
+
+impl<'a, K: AsRef<[u8]>, V: Serialize + 'a, It: Iterator<Item = (K, &'a Vec<V>)> + Clone>
+    LeavesStage<'a, K, V, It>
+{
+    pub(super) fn tree_stage(self) -> Result<TreeStage<'a, K, V, It>> {
+        // FIXME: remove collect (now it is used to iterate through `chunks` in `serialize_bptree`)
+        let keys: Vec<K> = self.headers_iter.clone().map(|e| e.0).collect();
```
pyakushin, line 143: Agree, `collect()` is redundant, you can pass `impl Iterator` into `serialize_bptree()`

src/blob/index/bptree/serializer.rs
```diff
@@ -0,0 +1,275 @@
...
+        let root_offset = Self::build_tree(leaf_nodes_compressed, tree_offset, &mut buf)?;
+        Ok((root_offset, buf))
+    }
+
+    pub(super) fn build_tree(
+        nodes_arr: Vec<(Vec<u8>, u64)>,
+        tree_offset: u64,
+        buf: &mut Vec<u8>,
+    ) -> Result<u64> {
+        if nodes_arr.len() == 1 {
```
pyakushin, line 194: I think you should check if `nodes_arr.is_empty()` also

src/blob/index/bptree/serializer.rs
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    fn max_leaf_node_capacity(key_size: usize) -> usize {
+        let offset_size = std::mem::size_of::<u64>();
+        BLOCK_SIZE / (key_size + offset_size) - 1
+    }
+
+    fn max_nonleaf_node_capacity(key_size: usize) -> usize {
+        let offset_size = std::mem::size_of::<u64>();
+        let meta_size = std::mem::size_of::<NodeMeta>();
```
pyakushin, line 222: Are you sure `size_of` is identical to `bincode::serialized_size()`?

src/blob/index/header.rs
pyakushin: I agree, we can use `u8` for both fields. 1 bit for `written` and 7 bits for version. I think we have enough versions reserved, 2^7 = 128 versions

pyakushin (commented):
2 part

src/blob/index/tools.rs
```diff
@@ -194,15 +11,30 @@ pub(crate) fn get_hash(buf: &[u8]) -> Vec<u8> {
 
 // if there is no elements, data will be wrong (because we can't get key_size),
 // BUT it will be computed correctly during first push
-pub(crate) fn compute_mem_attrs(record_headers: &InMemoryIndex) -> MemoryAttrs {
+pub(crate) fn compute_mem_attrs(
+    record_headers: &InMemoryIndex,
+    records_count: usize,
+) -> MemoryAttrs {
     let key_size = record_headers.keys().next().map_or_else(|| 0, |v| v.len());
```
pyakushin, line 18: I think you should return Err or panic if `key.len() == 0`

src/blob/index/bptree/core.rs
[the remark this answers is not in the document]
Justarone: @pyakushin I mean data write in active blob. During this operation `InMemoryIndex` will be updated (and will have at least 1 record header) so it can be successfully serialized. But if it hasn't at least one record: 
1. we are not able to determine key size; 
2. it's not worth to store file without record headers at all.

src/blob/index/bptree/core.rs
Justarone: @agend So, as I understood from your link, postgres stores per page checksum and during loading pages to shared buffer from disk it validates every loaded page. Postgres does that per page, because pages can be loaded independently. If modifications are done with any page (as I understand, they can be done only in shared buffer), checksum will be recalculated for this page when it will be removed from shared buffer (to OS cache or immediately to disk, it depends on behavior of OS and `fsync` usage in postgres implementation). Let's name postgres page as _exchange unit_ (portion of data, which is transfered between disk and shared buffer, and which can be modified (modification triggers checksum recalculation))

In our case, Pearl allows to load **all** key-record_header pairs in memory (`get_record_headers`; and then data can be modified, aka `shared buffer`) and write only **all** (at the same time I mean) key-record_header pairs (`from_records`; data can't be modified, aka `disk` but with read capability). So _exhange unit_  between InMemoryIndex (aka _Pearl shared buffer_) and disk is the whole index tree for blob. 

Did I misunderstand something? If I am not, I think, it's better leave checksum for the whole index tree, because we don't manipulate data (actually, we do, when we retrive elements array by key, but we don't modify them and we may validate checksums on startup for all disks (seems like it's not implemented yet, we only control hash during load in memory, so it is done only for active blob) and have guarantees, that it'll remain correct for all time) so if we'll add checksums for pages, we'll won't get any profit but instead we'll hit performance.

src/blob/index/bptree/core.rs
```diff
@@ -0,0 +1,283 @@
...
+    ) -> Result<Self> {
+        let res = Self::serialize(headers, filter)?;
+        let (mut header, metadata, buf) = res;
+        clean_file(path, recreate_index_file)?;
+        let file = File::create(path, ioring)
+            .await
+            .with_context(|| format!("file open failed {:?}", path))?;
+        file.write_append(&buf).await?;
+        header.written = 1;
+        let serialized_header = serialize(&header)?;
```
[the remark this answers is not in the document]
Justarone: That may lead to UB, if header structure will be changed (or serialization method of record header will be changed) our _byte write_ may become wrong.

src/blob/index/bptree/core.rs
Justarone: hm, there is unnecessary read of bptree, when we load record_headers (~ 25%), and actually, we can add second checksum for bptree (we validate it only at startup, and we don't need to validate it during retrieving of record headers). Is it what you meant?

src/blob/index/bptree/core.rs
```diff
@@ -0,0 +1,283 @@
...
+        path: &Path,
+        ioring: Option<Rio>,
+        headers: &InMemoryIndex,
+        filter: &Bloom,
+        recreate_index_file: bool,
+    ) -> Result<Self> {
+        let res = Self::serialize(headers, filter)?;
+        let (mut header, metadata, buf) = res;
+        clean_file(path, recreate_index_file)?;
+        let file = File::create(path, ioring)
```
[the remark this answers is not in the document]
Justarone: Actually, I don't create file 2 times (if file exists, then `clean_file` will truncate it, `File::create` will return descriptor for existing file). In `clean_file` I changed file deletion on truncation, because it's faster. Then I just recreate descriptor with our own file implemenation. If I'll just use `write_at(0, ...)`, seems like, I may get wrong bytes in tail.
But I missed case, when file doesn't exists in `clean_file`, changed that

src/blob/index/bptree/core.rs
```diff
@@ -0,0 +1,287 @@
...
+
+    async fn get_header_pointers(
+        &self,
+        leaf_offset: u64,
+        key: &[u8],
+        buf: &mut [u8],
+        leaf_size: usize,
+        buf_size: usize,
+    ) -> Result<Vec<(Vec<u8>, u64)>> {
+        let (buf, _rest) = buf.split_at_mut(buf_size as usize);
```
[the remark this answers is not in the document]
Justarone: I need a mutable reference on subslice, but your operation returns immutable one

src/blob/index/bptree/node.rs
Justarone: I mentioned this practice in std lib: for example, vector panics, if size is more than `isize::MAX`.

B+ Tree doesn't assume node to be without keys and children pointers amount has to be equal to `keys.len() + 1`. These are invariants, which can't be violated (So this state must be impossible). 
I don't violate them, so assert will never panic. But if we'll modify code, we may forget about it, so I decided to do it in a such way. But what are your suggestions?

src/blob/index/bptree/node.rs
Justarone: There is an error: this method doesn't exists for `Map<Chunks<..>`, because this struct doesn't implement `TryStream` trait (because it's an iterator; we can use `itertools` crate to use this method for iterator, but that's redundant imo)

src/blob/index/bptree/serializer.rs
Justarone: It was needed, because I wanted to extract the first element from iterator, but leave capability to extract it later again (for serialization).
This logic (with generalized serializer) is in #101 now, I removed it from this request, because generalization need to be discussed (and this serializer may affect performance)

src/blob/index/bptree/serializer.rs
Justarone: this case is impossible (there is no capability to pass 0 nodes on higher level of the tree), but I can make an assert. Should I?

src/blob/index/bptree/serializer.rs
Justarone: In this case - yes, but you're right, that it's better to use serialized_size, changed that

src/blob/index/tools.rs
Justarone: I'll panic

src/blob/index/bptree/core.rs
Justarone: sorry, I missed validation on startup:
While moving indices to disk, Pearl do 2 steps:
1. for serialized data it counts hash, writes this hash into header, and then writes header with serialized data into file
2. sets `written` byte in header, and writes new header in place of the old one

During startup `written` byte is checked (file still may be corrupted in case, when dirty page with written byte (with header, which also include right hash) was flushed to disk, but other pages - weren't; in this case written validation would be true, but hash validation would be false)

src/blob/index/header.rs
Justarone: Changed it, need your review!

src/blob/index/bptree/serializer.rs
Justarone: now keys are not collected

src/blob/index/simple.rs
```diff
@@ -1,377 +1,310 @@
...
-        if header.written != 1 {
-            return Err(Error::validation("header was not written").into());
-        }
-        if !Self::hash_valid(&header, &mut buf)? {
-            return Err(Error::validation("header hash mismatch").into());
-        }
-        if header.version != HEADER_VERSION {
-            return Err(Error::validation("header version mismatch").into());
+    fn validate(&self) -> Result<()> {
+        if self.header.written == 1 {
```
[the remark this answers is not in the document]
Justarone: As I mentioned above, this function only checks written byte because I need the whole file content to check hash (but when we start Pearl from file, we don't load that content, we load only header in RAM, so I check written byte in header in this case)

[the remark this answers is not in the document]
Justarone: I left logic, which was in simple index implementation

src/blob/index/bptree/core.rs
```diff
@@ -0,0 +1,287 @@
...
+        self.file.read_at(buf, leaf_offset).await?;
+        buf.chunks(leaf_size)
+            .map(|bytes| {
+                let (key, offset) = bytes.split_at(key.len());
```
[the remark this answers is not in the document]
Justarone: I am not sure about convenience, because in case of new `Leaf` struct I'll have to implement my own Serializer and Deserializer (because I don't need to store key vector size) + I need to pass key_size during serialization/deserialization, so I think this wrapper won't be more convenient imo

src/blob/index/bptree/core.rs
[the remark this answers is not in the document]
ikopylov: This is not critical, because dumping indexes in itself is a rare operation, and dumping indexes into an existing file is even rarer (occurs only in some exceptional cases)

src/blob/index/bptree/core.rs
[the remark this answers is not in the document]
Justarone: Yes, but if I'll rely on bincode serialization and write byte according to their implementation, there would be a trap for developer, and a bug can occur due to lib upgrade or something else
I don't think, that serializing and writing the whole header hits the performance, and IMO it's worth writing just one byte

If I misunderstood idea, give me a hint please

src/blob/index/bptree/core.rs
ikopylov: If there are `N` child nodes at each BPlusTree level, then the calculation will be: `1 + 1/N + 1/(N^2) + ...`. With `N = 16` result will be ~1.07. So there is no problem

src/blob/index/bptree/core.rs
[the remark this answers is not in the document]
ikopylov: On the first part: Yes, I meant `flush` operation. It is only applicable for `index` files.
> if dump is processed once a day, shouldn't we optimize this by doing this non-simultanously instead of changing fsync on flush?

It would be good optimization. How hard to do this?

On the second part: ok, I get it. This is a good optimization task, but I think it is better to do it in a separate request.

src/blob/index/header.rs
```diff
@@ -0,0 +1,86 @@
...
+            record_header_size,
+            meta_size,
+            hash,
+            ..Self::default()
+        }
+    }
+
+    pub(crate) fn set_written(&mut self, state: bool) {
+        self.version = self.version & !1;
+        self.version |= state as u8;
```
ikopylov, line 43: Please, recheck the conversion correctness. In C++ `true` `bool` value can be any non-negative number

src/blob/index/bptree/serializer.rs
```diff
@@ -0,0 +1,224 @@
...
+        // min_amount <= nodes left <= max_amount
+        let nodes_portion = &nodes_arr[current..];
+        Self::process_keys_portion(buf, nodes_portion, base_offset)?;
+        Ok(())
+    }
+
+    fn collect_next_layer_nodes(
+        nodes_arr: &[(Vec<u8>, u64)],
+        (min_amount, max_amount): (usize, usize),
+    ) -> Result<(Vec<(Vec<u8>, u64)>, u64)> {
```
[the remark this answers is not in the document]
ikopylov: An alias is even better. My concern was about readability, and aliases solve this problem 100%

src/blob/index/bptree/core.rs
```diff
@@ -153,29 +153,61 @@ impl BPTreeFileIndex {
...
 
+    fn get_leftmost(
+        &self,
+        raw_headers_buf: &[u8],
+        key: &[u8],
+        mut offset: usize,
+        mut prev_header: RecordHeader, // it's expected, that this header is from raw_headers_buf[offset..]
+        record_header_size: usize,
+    ) -> Result<RecordHeader> {
+        while offset > 0 {
```
ikopylov, line 181: Is it possible that there is no space before the current header for a one more header?
(Example: header size is 16 bytes and offset is 8 bytes)

src/blob/index/bptree/core.rs
```diff
@@ -0,0 +1,336 @@
...
+        record_header_size: usize,
+    ) -> Result<Option<(RecordHeader, usize)>> {
+        let mut l = 0i32;
+        let mut r: i32 = (headers.len() / record_header_size) as i32 - 1;
+        while l <= r {
+            let m = (l + r) / 2;
+            let m_off = record_header_size * m as usize;
+            let record_end = m_off + record_header_size;
+            let record_header: RecordHeader = deserialize(&headers[m_off..record_end])?;
+            match key.cmp(record_header.key()) {
```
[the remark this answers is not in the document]
ikopylov: > Do we still need to worry about this moment if we work in-memory?

From performance point of view, we don't. But the ordering issue is the global one, because if we change the order on disk (due to different key bytes interpretation) we will need to change it here too

src/blob/index/bptree/core.rs
```diff
@@ -0,0 +1,336 @@
...
+                CmpOrdering::Less => r = m - 1,
+                CmpOrdering::Greater => l = m + 1,
+                CmpOrdering::Equal => return Ok(Some((record_header, m_off))),
```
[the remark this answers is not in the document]
ikopylov: 👍

src/blob/index/bptree/core.rs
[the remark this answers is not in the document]
ikopylov: Current idea is to fix ordering by adding `PearlKey` with proper array comparison: https://github.com/qoollo/pearl/issues/123

src/blob/index/bptree/node.rs
```diff
@@ -0,0 +1,158 @@
...
+        Ok(meta_size + (keys_buf_size + offsets_buf_size) as u64)
+    }
+
+    pub(super) fn binary_search_serialized(key: &[u8], buf: &[u8]) -> Result<usize, usize> {
+        let mut l = 0i32;
+        let mut r: i32 = (buf.len() / key.len() - 1) as i32;
+        while l <= r {
+            let m = (l + r) / 2;
+            let offset = m as usize * key.len();
+            match key.cmp(&buf[offset..(offset + key.len())]) {
```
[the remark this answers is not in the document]
ikopylov: My comment was about changes in ordering due to different key bytes interpretation between Little and Big Endianness. Current idea is to fix this by adding `PearlKey` with proper array comparison: https://github.com/qoollo/pearl/issues/123

src/blob/index/bptree/node.rs
```diff
@@ -0,0 +1,158 @@
...
+            Ok(pos) => pos + 1,
+            Err(pos) => pos,
+        };
+        let offset = offsets_offset + ind * size_of::<u64>();
+        deserialize(&buf[offset..(offset + size_of::<u64>())]).map_err(Into::into)
+    }
+
+    #[allow(dead_code)]
+    pub(super) fn key_offset(&self, key: &[u8]) -> u64 {
+        match self.keys.binary_search_by(|elem| elem.as_slice().cmp(key)) {
```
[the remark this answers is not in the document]
ikopylov: Here I marked a place where the comparison works like the keys are in Big Endian format, but after recent changes they are in Little Endian format. Current idea is to fix ordering by adding `PearlKey` with proper array comparison: #123

src/blob/index/bptree/core.rs
Justarone: @ikopylov On practice there are some nuances (alignment of root node, meta of every node, tree meta also need some place) BUT I remembered that my calculations are related to version, when leaves weren't headers themselves, but 'pointers' (offset in file) on headers, so at current version amount of additional size is reduced much (and I guess it will be no more than 7%, as you calculated)
If you are interested in exact value please write about this, otherwise resolve this conversation

src/blob/index/bptree/core.rs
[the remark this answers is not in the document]
Justarone: Seems like this task is better to do in separate PR, because this one is already too big.. Resolving this conversation

src/blob/index/bptree/node.rs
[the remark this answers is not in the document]
Justarone: Same here:

> Seems like this task is better to do in separate PR, because this one is already too big.. Resolving this conversation

src/blob/index/bptree/node.rs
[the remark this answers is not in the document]
Justarone: Same here:

> Seems like this task is better to do in separate PR, because this one is already too big.. Resolving this conversation

src/blob/index/bptree/core.rs
Justarone: No, it's not possible, because it's guaranteed that every leaf node starts from header (and even more: it's guaranteed, that there is no situation, that first header in node has the same key, as previous header) and we read first 4kb of node in `raw_headers_buf`

src/blob/index/header.rs
Justarone: It is guaranteed [here](https://doc.rust-lang.org/std/primitive.bool.html), but if you worry that one day this could be changed I can match bool value and set 1 explicitly (otherwise resolve conversation please)

src/blob/index/bptree/core.rs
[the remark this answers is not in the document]
Justarone: Change fsync on flush for headers.
> It would be good optimization. How hard to do this?
Pearl is only responsible for closing blobs on bytes/records limit excess. 

Bob is responsible for closing blob due to timestamp excess, so this work should be done in Bob crate (it doesn't looks like a difficult task: it would be enough to append offset to timestamp start (and end) value(-s) according, for example, to vdiskid)

src/blob/index/simple.rs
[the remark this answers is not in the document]
Justarone: Added FIXME: as this is a part of old logic I'll leave it like this, but you'll have a notice, that maybe this should be changed.

ikopylov:
Please, fix broken tests and update CHANGELOG.md.

Also, when https://github.com/qoollo/pearl/pull/102 will be merged, the line about Range indices should be added to CHANGELOG.md

pyakushin (approved):
LGTM

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@ Pearl changelog
 
 ## [Unreleased]
 #### Added
+- B+ Tree indexes (#84)
```
ikopylov, line 6: Please, add a line about Range Indices here.

Additionally: it is better to write a link to the Pull Request in braces.

src/blob/index/bptree/core.rs
```diff
@@ -0,0 +1,369 @@
...
+
+    async fn get_any(&self, key: &[u8]) -> Result<Option<RecordHeader>> {
+        let root_offset = self.metadata.tree_offset;
+        let mut buf = [0u8; BLOCK_SIZE];
+        let leaf_offset = self.find_leaf_node(key, root_offset, &mut buf).await?;
+        self.read_header(leaf_offset, key, &mut buf).await
+    }
+
+    fn validate(&self) -> Result<()> {
+        // FIXME: check hash here?
```
ikopylov, line 124: Probably, index file version should also be checked here

ikopylov (approved):
LGTM

PR #94, opened 2021-04-01
Rebuild corrupted index automatically during startup

### LINKED ISSUE #88: Rebuild corrupted index file automatically during startup
Related to https://github.com/qoollo/bob/issues/191

vovac12:
Closes #88


src/blob/core.rs
```diff
@@ -112,9 +112,15 @@ impl Blob {
         let mut index_name = name.clone();
         index_name.extension = BLOB_INDEX_FILE_EXTENSION.to_owned();
         trace!("looking for index file: [{}]", index_name);
+        let mut is_index_corrupted = false;
         let index = if index_name.exists() {
             trace!("file exists");
-            SimpleIndex::from_file(index_name, filter_config.is_some(), ioring).await?
+            SimpleIndex::from_file(index_name.clone(), filter_config.is_some(), ioring.clone())
```
pyakushin, line 118: I think it's better to use `if let Err(e) = ...` to skip index regeneration in case of some IO errors. Plus some logs will be helpful.

pyakushin (changes requested):

src/blob/core.rs
vovac12: I added skipping index regeneration on `PermissionDenied` and `Other` IO errors.

ikopylov:
LGTM

pyakushin (approved):

PR #98, opened 2021-04-14
Move corrupted blob into special dir

### LINKED ISSUE #89: Move corrupted BLOBs into a special directory
This should allow manual recovery later

vovac12:
Closes #89


pyakushin (approved):
LGTM
There are a couple suggestions though

src/blob/core.rs
```diff
@@ -214,6 +228,18 @@ impl Blob {
         debug!("blob read any entry found");
         let buf = entry
             .load()
+            .then(|result| async {
+                let error = match result {
```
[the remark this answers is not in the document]
vovac12: I can't use await inside sync closure. 
I can avoid of `match` with this code
```rust
.then(|result| async {
  if let Err(error) = &result {
    if let Some(error) = error.downcast_ref::<Error>() {
      if let ErrorKind::Validation(_) = error.kind() {
        self.dump_corrupted().await?;
      }
    }
  }
  result
})
```

src/blob/core.rs
```diff
@@ -153,6 +153,22 @@ impl Blob {
             .context("failed to create iterator for raw records")
     }
 
+    async fn dump_corrupted(&self) -> Result<()> {
+        warn!("Blob {} dump started!", self.name());
+        let source_name = self.name.to_path();
+        let file_name = source_name
+            .file_name()
+            .ok_or_else(|| anyhow::anyhow!("Blob path is empty"))?;
+        let dir = self.name.dir.join("corrupted");
```
ikopylov, line 162: Hardcoding this name is not good.  Declare public constant or get this name from Pearl config

src/blob/core.rs
```diff
@@ -153,6 +153,22 @@ impl Blob {
             .context("failed to create iterator for raw records")
     }
 
+    async fn dump_corrupted(&self) -> Result<()> {
+        warn!("Blob {} dump started!", self.name());
```
Justarone, line 157: Pearl users may be confused reading this message, I think we should also notice, that the blob is corrupted.

src/blob/core.rs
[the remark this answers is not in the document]
Justarone: I think current version is better

vovac12:
I moved corrupted blobs processing to `Storage`. IMO it's better solution

src/storage/core.rs
```diff
@@ -518,6 +536,47 @@ impl<K: Key> Storage<K> {
...
+
+    async fn save_corrupted_blob(path: &Path, corrupted_dir_name: &str) -> Result<()> {
+        let parent = path.parent().ok_or_else(|| {
+            anyhow::anyhow!("[{}] blob path don't have parent directory", path.display())
+        })?;
+        let file_name = path
+            .file_name()
+            .ok_or_else(|| anyhow::anyhow!("[{}] blob path don't have file name", path.display()))?
+            .to_os_string();
+        let corrupted_path = parent.join(corrupted_dir_name).join(file_name);
```
ikopylov, line 557: Here we have lost the directory level with a timestamp, so it will be impossible to properly recover corrupted BLOBs. You have to add the name of the directory inside which the blob is placed: 
`let corrupted_path = parent.join(corrupted_dir_name).join(cur_blob_dir).join(file_name);`

ikopylov (changes requested):
Fixes are required

src/storage/core.rs
vovac12: It's ok here. `parent` is directory with blobs. `path` argument contains path to blob.

ikopylov (approved):
LGTM

PR #99, opened 2021-04-14
Add key size check for blob from file

Justarone:
During index regeneration operations from blob file, first record of blob file is checked to have key of the same size, as compile-time key size for Pearl storage (if not equal - validation error).


idruzhitskiy (commented):
LGTM

PR #101, opened 2021-04-19
B+ Tree generalization

Justarone:
We have in plans a move of our own b+ tree file implementation in separate crate, so first we need to make it self-sufficient and independent (generalize it, because if it will only work with our records it won't be useful enough). I highlight 3 steps:

- [x] Make first generalized version of b+ tree
- [ ] Fix weaknesses and bugs of first version
- [ ] Reorgonize module structure finally to get ready for the move (move all structs (such as `File`, `IndexHeader`) on which bptree depends closer to bptree module)

During solving the 1 step I encountered some difficulties, which means, that second step won't be instantaneous. 
Changes are significant AND I am not sure, that I will do it in right way from the first attempt, so to prevent reverts and weird manipulations, I decided to separate generalization in another PR and work on generalization in it.

### DIFF
--- a/src/blob/index/bptree/core.rs
+++ b/src/blob/index/bptree/core.rs
@@ -13,19 +13,26 @@
 ///    `NodeMeta | keys_arr | pointers_arr`,
 ///    where pointer is offset in file of underlying node with searched key
 use super::prelude::*;
+use serde::{de::DeserializeOwned, Serialize};
 
 pub(super) const BLOCK_SIZE: usize = 4096;
 
 #[derive(Debug, Clone)]
-pub(crate) struct BPTreeFileIndex {
+pub(crate) struct BPTreeFileIndexStruct<K, V> {
     file: File,
     header: IndexHeader,
     metadata: TreeMeta,
+    kmarker: PhantomData<K>,
+    vmarker: PhantomData<V>,
 }
 
-#[async_trait::async_trait]
-impl FileIndexTrait for BPTreeFileIndex {
-    async fn from_file(name: FileName, ioring: Option<Rio>) -> Result<Self> {
+impl<
+        'a,
+        K: 'a + AsRef<[u8]> + Ord + DeserializeOwned + Clone,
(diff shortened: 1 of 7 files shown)


Justarone:
Generalization is in progress, so I think it's better to draft this request for this moment (I plan to solve this task this week but it depends on the complexity)

Justarone:
@pyakushin , @idruzhitskiy , @agend I need your advice.

Generalization task appeared pretty complicated for me: it's difficult to cope with those generic bounds (and it may take much time to implement it)... So there are several advantages and disadvantages of this generalization work.

Advantages:
* Useful self-sufficient crate with generic b+tree serializer, deserializer and file manager (search capability)

Disadvantages:
* Generic code complexity
    * Harder to implement (I have broken my head already)
    * Harder to maintain (little changes may be tricky to implement)
    * Harder to understand (main logic would be partially hidden by generic work code)
* Performance hit OR more complicated code with stricter generic bounds

I also see one more option, how to simplify generic code (but with performance hit on serialization process): We can implement it for keys and values presented as `Vec<u8>` and move serialization/deserialization work on requester side. But there would be weird chain (for example, for Pearl: `File -> BTreeMap<Vec<u8>, Vec<u8>> -> BTreeMap<Key, RecordHeader>`). Pearl is performance sensitive, so I think, that's not okay for Pearl, is it?

So I guess, we should postpone this task (generalization of bptree and its move to separate crate) for a while, and firstly we need to:

- [ ] benchmark `BPtree` and compare it to `Simple` (which was used before)
    [I'll do that in short time, I hope]
- [ ] come to the detailed solution about generic bounds and other tricky parts of generalization tasks (how data is passed to bptree (BTreeMap or Iterator or etc), how data is retrieved from it and etc)
    I have already discussed this a bit with @idruzhitskiy , but I encountered some difficulties during implementation.
- [ ] discuss generic implementation restrictions, complexity and possible performance hit (and decide, is it worth to implement it)

Issue #11, opened 2019-04-03
Implement Range index for BLOB


ikopylov:
It would allow to reject not presented entries faster

PR #102, opened 2021-05-26
Add range indices

Justarone:
As I already mentioned in code, this version is inefficient for keys represented in little endian format, because vectors are compared from left to right (so from less grade to higher one) so, for example if we have u16 keys:
In worst case if range filter contains only numbers between 0 and 255 and these numbers (which are represented as [0, 0] and [255, 0]) there are only 255 numbers which will be filtered (511 = [255, 1], 767 = [255, 2], ... , 65535 = [255, 255]), but actually we can filter more than 65k (255, 256, 257, ... , 65535)

Default serializers (for example in [`Bob`](https://github.com/qoollo/bob) crate, which uses `Pearl`, [`bincode`](https://github.com/bincode-org/bincode) crate is used) for numbers try to work efficiently so they just copy numbers representation in memory to vectors and it [representation] depends on arch (little/big endian)

I have 3 solutions:
1. make it work only for little endian (because most of modern CPU uses little endian) and mention that somewhere (so for big endian systems it would be better to write custom serializer)
2. make it work for both little and big endian (I think there won't be performance and memory overheads)
3. make an optional field in config to choose, how to compare vector-keys (straight or reversed order; by default - reversed (check [1], why it is better))

### DIFF
--- a/src/blob/core.rs
+++ b/src/blob/core.rs
@@ -328,7 +328,11 @@ impl Blob {
 
     pub(crate) fn check_filters(&self, key: &[u8]) -> bool {
         trace!("check filters (range and bloom)");
-        self.index.check_filters_key(key)
+        if let FilterResult::NotContains = self.index.check_filters_key(key) {
+            false
+        } else {
+            true
+        }
     }
 
     pub(crate) async fn check_filters_non_blocking(&self, key: &[u8]) -> bool {
--- a/src/blob/core.rs
+++ b/src/blob/core.rs
@@ -215,10 +215,15 @@ impl Blob {
         Ok(())
     }
 
-    pub(crate) async fn read_any(&self, key: &[u8], meta: Option<&Meta>) -> Result<Vec<u8>> {
+    pub(crate) async fn read_any(
+        &self,
+        key: &[u8],
+        meta: Option<&Meta>,
+        check_filters: bool,
+    ) -> Result<Vec<u8>> {
         debug!("blob read any");
         let entry = self
-            .get_entry(key, meta)
+            .get_entry(key, meta, check_filters)
             .await?
             .ok_or_else(|| Error::from(ErrorKind::RecordNotFound))?;
         debug!("blob read any entry found");
@@ -247,10 +252,15 @@ impl Blob {
             .collect()
     }
 
-    async fn get_entry(&self, key: &[u8], meta: Option<&Meta>) -> Result<Option<Entry>> {
+    async fn get_entry(
+        &self,
+        key: &[u8],
+        meta: Option<&Meta>,
+        check_filters: bool,
+    ) -> Result<Option<Entry>> {
         debug!("blob get any entry {:?}, {:?}", key, meta);
-        if self.check_filters(key) == Some(false) {
-            debug!("blob core get any entry check bloom returned Some(false)");
+        if check_filters && !self.check_filters(key) {
+            debug!("Key was filtered out by filters");
             Ok(None)
         } else if let Some(meta) = meta {
             debug!("blob get any entry meta: {:?}", meta);
@@ -293,7 +303,7 @@ impl Blob {
 
     pub(crate) async fn contains(&self, key: &[u8], meta: Option<&Meta>) -> Result<bool> {
         debug!("blob contains");
-        let contains = self.get_entry(key, meta).await?.is_some();
+        let contains = self.get_entry(key, meta, true).await?.is_some();
         debug!("blob contains any: {}", contains);
         Ok(contains)
     }
@@ -316,17 +326,13 @@ impl Blob {
         self.name.id
     }
 
-    pub(crate) fn check_filters(&self, key: &[u8]) -> Option<bool> {
+    pub(crate) fn check_filters(&self, key: &[u8]) -> bool {
         trace!("check filters (range and bloom)");
         self.index.check_filters_key(key)
     }
 
-    pub(crate) async fn check_filters_async<R>(&self, key: &[u8], index: R) -> Option<R> {
-        if self.check_filters(key).unwrap_or(true) {
-            Some(index)
-        } else {
-            None
-        }
+    pub(crate) async fn check_filters_non_blocking(&self, key: &[u8]) -> bool {
+        self.check_filters(key)
     }
 
     pub(crate) fn index_memory(&self) -> usize {
--- a/src/blob/core.rs
+++ b/src/blob/core.rs
@@ -321,6 +321,14 @@ impl Blob {
         self.index.check_filters_key(key)
     }
 
+    pub(crate) async fn check_filters_async<R>(&self, key: &[u8], index: R) -> Option<R> {
+        if self.check_filters(key).unwrap_or(true) {
+            Some(index)
+        } else {
+            None
+        }
+    }
+
     pub(crate) fn index_memory(&self) -> usize {
         self.index.memory_used()
     }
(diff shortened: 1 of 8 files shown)


Justarone:
closes #11

Justarone:
Problem solved: I set compile-time flag, so depending on platform that would be chosen we'll have different checks (which in cases of sequential keys writes will filter move keys, check the example above)

Justarone:
in related crate [`bob`](https://github.com/qoollo/bob) we use `to_be_bytes` method to convert u64 to bytes and it converts in big-endian format, so I decided to leave big-endian format comparison for range indices BUT I left a NOTE for the future, if keys format will be changed (the problem is that Pearl doesn't really know in which format (big/little-endian) library above  (which uses Pearl, such as Bob) will pass keys)

Justarone:
I also add implementation which returns last record for searched key (parallel bloom filter + `FuturesOrdered` search in blobs' index-files and in blobs themselves)

src/blob/core.rs
```diff
@@ -316,9 +316,17 @@ impl Blob {
...
 
-    pub(crate) fn check_bloom(&self, key: &[u8]) -> Option<bool> {
-        trace!("check bloom filter");
-        self.index.check_bloom_key(key)
+    pub(crate) fn check_filters(&self, key: &[u8]) -> Option<bool> {
+        trace!("check filters (range and bloom)");
+        self.index.check_filters_key(key)
+    }
+
+    pub(crate) async fn check_filters_async<R>(&self, key: &[u8], index: R) -> Option<R> {
```
idruzhitskiy, line 324: It's a bit unintuitive to see `index: R` here... The only reason for this is `map` in `get_data_last`? Probably it's better to let it remain simple predicate and use `filter`?

src/storage/core.rs
```diff
@@ -303,7 +304,31 @@ impl<K: Key> Storage<K> {
         }
     }
 
-    async fn get_any_data(safe: &Safe, key: &[u8], meta: Option<&Meta>) -> Result<Vec<u8>> {
+    async fn get_data_last(safe: &Safe, key: &[u8], meta: Option<&Meta>) -> Result<Vec<u8>> {
+        let blobs = safe.blobs.read().await;
+        let blobs_stream: FuturesOrdered<_> = blobs
```
idruzhitskiy, line 309: Looks like blob has filters check inside read_any ( https://github.com/qoollo/pearl/blob/75e5c8793216138c639d95f60ad66251b2606938/src/blob/core.rs#L252 ). So you don't need another check here

src/blob/index/core.rs
```diff
@@ -81,12 +84,15 @@ impl<FileIndex: FileIndexTrait> IndexStruct<FileIndex> {
...
-        self.filter.clear();
+        self.bloom_filter.clear();
+        self.range_filter.clear();
     }
 
-    pub fn check_bloom_key(&self, key: &[u8]) -> Option<bool> {
-        if self.params.filter_is_on {
-            Some(self.filter.contains(key))
+    pub(crate) fn check_filters_key(&self, key: &[u8]) -> Option<bool> {
+        if !self.range_filter.contains(key) {
```
ikopylov, line 92: There is an error in this condition chain. If `range_filter` returns `true` and bloom filter is off, the result should be `true`, not `None`

src/blob/index/range.rs
```diff
@@ -0,0 +1,60 @@
...
+        let key_slice = key.as_ref();
+        if !self.initialized {
+            self.min = key_slice.to_vec();
+            self.max = key_slice.to_vec();
+            self.initialized = true;
+        } else if Self::lt(key_slice, &self.min) {
+            self.min = key_slice.to_vec()
+        } else if Self::lt(&self.max, key_slice) {
+            self.max = key_slice.to_vec()
+        }
```
ikopylov, line 29: Is there a guarantee, that this code will be called from a single thread? Because it is not thread safe

src/storage/core.rs
ikopylov: > Core idea was to check indices concurrently and then check consistently all blobs, that are not filtered out by indices

This is correct

>should we use FuturesUnordered (check asyncronously, but analyzing results according to order) or check really consistently

The result of this check should be ordered (this is critical for correctness). But this checks should be done concurrently for better perf

src/blob/index/core.rs
Justarone: Actually, there is no, but I rewrite this function a bit and now it returns just bool.
Previous version returned:
1) None (if key can't be checked due to bloom filter off state) aka 'no information, go and check'
2) Some(true) (if key exists according to bloom filter (it means "go and check explicitly")) aka 'likely exists but check'
3) Some(false) (if key doesn't exist **for sure**) aka 'not for sure'

Now I decided to unite None and Some(true) variants and return just false (doesn't exist for sure) and true (maybe exists)

This code is okay and here's it's logic:
1. if range filters **doesn't** contains this key (aka 'is key < min or key > max?') => return Some(false) aka 'not for sure'
2. if bloom is on then return Some(bloom_result), where bloom_result either true (=> then return Some(true) aka 'likely exists but check') or false (=> then return Some(false) aka 'not for sure')
3. if bloom is off then return None aka 'no information, go and check' (Some(true) aka 'likely exists but check' also can be returned in this case, but it doesn't matter)

So to not confuse someone else I made it return just true or false.

src/blob/index/range.rs
Justarone: Yes, there is.
Simple answer: Rust allows only 1 mutable ref **or** N shared (immutable) refs (and doesn't allow to have it simultanously) and if you don't use `unsafe` then this guarantee won't be violated (and we don't). 
More strict answer: `Safe` struct which contains all blobs is under `RWLock`, so this is a guarantee.

src/blob/core.rs
Justarone: Okay, I was bitten by a crazy frog (or I don't know why I dicided to write it in that way), so now I did in a right way

src/storage/core.rs
Justarone: @idruzhitskiy I was confused by 2 simultaneous problems, which I solve here so thanks for found issue, resolved it

src/blob/index/core.rs
```diff
@@ -81,14 +84,17 @@ impl<FileIndex: FileIndexTrait> IndexStruct<FileIndex> {
...
         self.mem = Some(Default::default());
-        self.filter.clear();
+        self.bloom_filter.clear();
+        self.range_filter.clear();
     }
 
-    pub fn check_bloom_key(&self, key: &[u8]) -> Option<bool> {
-        if self.params.filter_is_on {
-            Some(self.filter.contains(key))
+    pub(crate) fn check_filters_key(&self, key: &[u8]) -> bool {
```
ikopylov, line 91: For readability it is better to introduce `enum` with `NeedAdditionalCheck` and `NotContains` variants. `bool` is not the best type here

ikopylov:
One small comment, apart from that LGTM

ikopylov:
LGTM

@pyakushin please, take a look

src/storage/core.rs
```diff
@@ -303,10 +304,42 @@ impl<K: Key> Storage<K> {
...
+            .enumerate()
+            .map(|(i, blob)| async move {
+                if blob.check_filters_non_blocking(key).await {
+                    Some(i)
+                } else {
+                    None
+                }
+            })
+            .collect();
+        let possible_blobs: Vec<_> = blobs_stream.filter_map(|e| e).collect().await;
```
pyakushin, line 320: Why you didn't use `filter_map` on L312?

Justarone: Because closure in that `map` returns `future` (async block), but `filter_map` expects `Option`.
Resolve conversation if the answer is ok.

PR #103, opened 2021-06-02
Mark records as deleted

vovac12:


src/storage/core.rs
```diff
@@ -315,6 +315,46 @@ impl<K: Key> Storage<K> {
...
+                debug!(
+                    "storage core mark all as deleted closed blob count {}",
+                    count
+                );
+                total_count += count;
+                future::ok(())
+            })
+            .await?;
+        debug!(
+            "storage core mark all as deleted total count {}",
```
idruzhitskiy, line 352: You can setup logging to output file and position ( https://docs.rs/log/0.4.14/log/struct.Record.html#method.file ), if you need, so you should remove file marks from log messages.

src/blob/index/simple.rs
```diff
@@ -278,6 +278,28 @@ impl Simple {
             0
         }
     }
+
+    fn mark_all_as_deleted_in_memory(&mut self, key: &[u8]) -> Result<Option<u64>> {
```
idruzhitskiy, line 282: IMO it's better to omit "_in_memory" in the name of the function, as it's implementation detail, as this function deals with both cases. And it's better to extract inner logic of headers cleanup into separate function, so match branches are smaller and it's easier to read

src/storage/core.rs
```diff
@@ -315,6 +315,46 @@ impl<K: Key> Storage<K> {
             .with_context(|| "no results in closed blobs")
     }
 
+    /// Mark as deleted entries with matching key
+    /// # Errors
+    /// Fails after any disk IO errors.
+    pub async fn mark_all_as_deleted(&self, key: impl AsRef<K>) -> Result<u64> {
+        let mut total_count = 0u64;
+        let mut safe = self.inner.safe.write().await;
```
idruzhitskiy, line 323: You should split this function into `mark_all_as_deleted_active` and `mark_all_as_deleted_closed`, so it's intentions are easier to find out

src/storage/core.rs
vovac12: Fixed

src/blob/index/simple.rs
vovac12: Done

src/storage/core.rs
vovac12: Done

idruzhitskiy:
LGTM

src/blob/core.rs
```diff
@@ -244,6 +245,25 @@ impl Blob {
         }))
     }
 
+    #[inline]
```
pyakushin, line 248: I think it's redundant here, because won't add any performance, but will increase compilation time

src/storage/core.rs
```diff
@@ -315,6 +315,49 @@ impl<K: Key> Storage<K> {
...
+            .collect::<FuturesUnordered<_>>();
+        let total = entries_closed_blobs
+            .filter_map(|result| match result {
+                Ok(count) => count,
+                Err(error) => {
+                    warn!("failed to delete records: {}", error);
+                    None
+                }
+            })
+            .fold(0, |a, b| a + b)
```
pyakushin, line 344: You can use `.sum()`

src/blob/index/simple.rs
```diff
@@ -278,6 +278,17 @@ impl Simple {
             0
         }
     }
+
+    fn mark_all_as_deleted_in_memory(headers: &mut InMemoryIndex, key: &[u8]) -> Option<u64> {
+        debug!("headers: {}", headers.len());
+        if let Some(vec_headers) = headers.get_mut(key) {
```
idruzhitskiy, line 284: I think it's better to return directly `headers.get_mut(key).map(|vec_headers| { ... })`

idruzhitskiy (approved):

src/blob/core.rs
```diff
@@ -517,6 +536,26 @@ impl RawRecords {
     }
 }
 
+pub(crate) fn filter_deleted_headers(headers: Vec<RecordHeader>) -> Vec<RecordHeader> {
```
ikopylov, line 539: Deletion is a rare operation, so the algorithm can be improved for the common case:
you can collect in BTreeMap the latest deleted header only for keys, that was marked as deleted. If this map becomes empty, then it is safe to return initial `headers` vector. Otherwise, you filter original records by condition: `header_map` not contains key or `header.created()` is large then last deletion time for the key

src/blob/core.rs
```diff
@@ -517,6 +536,27 @@ impl RawRecords {
...
+        }
+        map
+    });
+    if deleted.is_empty() {
+        return headers;
+    }
+
+    let mut new_headers = vec![];
+    for header in headers {
+        if deleted.get(header.key()).cloned().unwrap_or_default() < header.created() {
```
ikopylov, line 553: What happens when the key is not present in `BTreeMap`? 
In this case, the header should normally be copied into `new_headers`.

vovac12: In this case condition will look like `if 0 < header.created()` and this is always true for valid timestamps.

ikopylov (approved):
LGTM

PR #105, opened 2021-06-07
Capability to perform update operation

Justarone:
Update search strategy:
1. Bloom is checked as FuturesOrdered (to get indices of possibly useful blobs in ordered format, this doesn't affect performance because we need to wait for all bloom checks anyway)
2. Bloom-true blobs checked as FuturesOrdered in reversed order and the first (last === first in reversed order) is returned

This allows to retrieve the last record ==> define `update` operation as `write new` operation.

I checked this strategy and it doesn't affect performance much (but it depends on configuration (on my configuration it didn't affect at al), so I think it's better to leave both implementations)

### DIFF
--- a/src/blob/core.rs
+++ b/src/blob/core.rs
@@ -215,10 +215,15 @@ impl Blob {
         Ok(())
     }
 
-    pub(crate) async fn read_any(&self, key: &[u8], meta: Option<&Meta>) -> Result<Vec<u8>> {
+    pub(crate) async fn read_any(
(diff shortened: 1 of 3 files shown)


Justarone:
IMO `check_bloom_async` looks weird, I need some hints how to do it better

Justarone:
Changes from this PR exist in [PR with range indexes](https://github.com/qoollo/pearl/pull/102)

PR #110, opened 2021-07-08
108 platform agnostic

### LINKED ISSUE #108: Do not use platform dependent `usize` for serializable data
This can lead to binary incompatibility between 32 and 64 bit version. All data sizes must be specified explicitly using `u64`. 

https://github.com/qoollo/pearl/blob/master/src/blob/index/simple.rs#L19
https://github.com/qoollo/pearl/blob/master/src/blob/index/bloom.rs#L14

idruzhitskiy:
closes #108


ikopylov:
LGTM

ikopylov:
Here is the incorrect assumption about `usize`: https://github.com/qoollo/pearl/blob/master/src/blob/core.rs#L431
Please, fix it in this request

pyakushin (approved):

PR #111, opened 2021-07-20
New bits count calculations for bloom filter

Justarone:
New method calculates bits count values directly instead of iterations

### DIFF
--- a/src/blob/index/bloom.rs
+++ b/src/blob/index/bloom.rs
@@ -36,7 +36,7 @@ struct Save {
 impl Default for Config {
     fn default() -> Self {
         Self {
-            elements: 1_000_000,
+            elements: 100_000,
             hashers_count: 2,
             max_buf_bits_count: 8_388_608, // 1Mb
             buf_increase_step: 8196,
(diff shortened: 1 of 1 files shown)


src/blob/index/bloom.rs
```diff
@@ -45,32 +45,54 @@ impl Default for Config {
...
+
+fn bits_count_from_formula(config: &Config) -> usize {
+    let max_bit_count = config.max_buf_bits_count; // 1Mb
+    trace!("max bit count: {}", max_bit_count);
+    let k = config.hashers_count;
+    let n = config.elements as f64;
+    trace!("bloom filter for {} elements", n);
+    let mut bits_count = (n * k as f64 / 2_f64.ln()) as usize;
+    let fpr = config.preferred_false_positive_rate;
+    bits_count = bits_count.max(max_bit_count.min(m_from_fpr(fpr, k as f64, n) as usize));
```
ikopylov, line 60: From config we can setup only `max_bit_count`, so i suggest to prefer its value: 
```
bits_count = max_bit_count.min(bits_count.max(m_from_fpr(fpr, k as f64, n) as usize));
```

pyakushin (approved):

PR #112, opened 2021-07-27
246 broken data compatibility

pyakushin:


Justarone:
I finally found rpm package, on which broken compatibility was found and... it still exists. As I remember we have discussed 2 solutions:
1. Revert library version
2. Move code from library to get rid of this dangerous dependency and control it ourselves

I see, that you have fixed some parameters to prevent this kind of problems in the future (and in this case code looks good to me (maybe we also should fix crc version "=2.0.0")), but it doesn't solve existing problem. @ikopylov , need your advice.

Justarone:
And I don't think that it's a good idea to update tokio version here either, because if it will break something here, we will revert breaking changes, but we'll revert fixing changes (for this problem) also

ikopylov:
We decided to break backward compatibility and freeze the library version at the newest one. The reason is that the next major release should also contain BPlusTree index (https://github.com/qoollo/pearl/pull/93), which is also a breaking change, so we allow it here as well. 

@pyakushin But it is still important to increment index file version.

> maybe we also should fix crc version "=2.0.0"

CRC is a super common algorithm, so the calculation can only change due to the bug in the library. However, it would be nice to write a test, that checks CRC calculation.

src/lib.rs
```diff
@@ -70,14 +70,16 @@ pub use rio;
 pub use storage::{Builder, Key, Storage};
 
 mod prelude {
+    pub const CASTAGNOLI: Crc<u32> = Crc::<u32>::new(&CRC_32_ISCSI);
```
idruzhitskiy, line 73: IMO it's better to name it CRC32C

tests/tests.rs
```diff
@@ -19,6 +20,17 @@ mod common;
 
 use common::KeyTest;
 
+#[test]
+fn test_hash() {
+    let data = (0..10).collect::<Vec<u8>>();
```
idruzhitskiy, line 25: Maybe there should be multiple cases, 3-5, including 0 and something large

src/lib.rs
pyakushin: Ok

pyakushin:
> And I don't think that it's a good idea to update tokio version here either, because if it will break something here, we will revert breaking changes, but we'll revert fixing changes (for this problem) also

1. I double checked, there are no breaking changes. 
2. But fixed a couple of memory leaks

Justarone (approved):

PR #118, opened 2021-08-12
Extend interface and make active blob fully optional

### LINKED ISSUE #117: Extend interface to add capability to work without active blob
Related to [issue in other project](https://github.com/qoollo/bob/issues/285)

Justarone:
closes #117

### DIFF
--- a/src/storage/core.rs
+++ b/src/storage/core.rs
@@ -109,6 +109,25 @@ impl<K: Key> Storage<K> {
     ///
     /// [`init()`]: struct.Storage.html#method.init
     pub async fn init(&mut self) -> Result<()> {
+        self.init_ext(true).await
+    }
+
+    /// [`init_noactive()`] used to prepare all environment to further work, but unlike `init`
+    /// doesn't set active blob, which means that write operations will fall with
+    /// `active_blob_not_set`.
+    ///
+    /// Storage works in directory provided to builder. If directory don't exist,
+    /// storage creates it, otherwise tries to init existing storage.
+    /// # Errors
+    /// Returns error in case of failures with IO operations or
+    /// if some of the required params are missed.
+    ///
+    /// [`init_noactive()`]: struct.Storage.html#method.init
+    pub async fn init_noactive(&mut self) -> Result<()> {
+        self.init_ext(false).await
+    }
+
+    async fn init_ext(&mut self, with_active: bool) -> Result<()> {
         // @TODO implement work dir validation
         self.prepare_work_dir().await?;
         let cont_res = work_dir_content(
@@ -121,7 +140,7 @@ impl<K: Key> Storage<K> {
         trace!("work dir content loaded");
         if let Some(files) = cont_res? {
             trace!("storage init from existing files");
-            self.init_from_existing(files)
+            self.init_from_existing(files, with_active)
                 .await
                 .context("failed to init from existing blobs")?
         } else {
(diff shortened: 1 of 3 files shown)


src/storage/core.rs
```diff
@@ -177,6 +261,10 @@ impl<K: Key> Storage<K> {
         meta: Option<Meta>,
     ) -> Result<()> {
         debug!("storage write with {:?}, {}b, {:?}", key, value.len(), meta);
+        // if active blob is set, this call will only check this fact and return
+        if self.create_active_blob().await? {
```
ikopylov, line 265: Non-obvious behavior for a function with this name. It is better to rename function to `create_active_blob_if_not_exist`

ikopylov:
Can you re-check 2 sceanrious:
1. What happens if initialization with an active BLOB is specified, but there is no BLOB in the directory? 
2. What happens if an active BLOB is restored, but it was finalized and has an index file on disk? When the index is dumped, will it override the existing one?

Justarone:
@ikopylov 
1. I use the scenario which was used before: if there are 1+ blobs in a directory, blob with the greatest id will become active (its index will be loaded in memory from file and after that this index will be dumped, rewriting content of index file), otherwise new blob will be created
2. Some kind of `restore` operation already was used, when Pearl is initialized from non-empty storage. As I mentioned higher, it gets highest-id blob, load it in memory, and when this blob big enough (data or elements) it is dumped by observer (and it rewrites index).

src/storage/core.rs
Justarone: This behavior is defined by return type `Result<bool>`. If an error occurs, then `Result::Err(e)` is returned, otherwise `Result::Ok(true)` is returned in case, when active blob was really created, `Result::Ok(false)` - when it wasn't. I mentioned this in docs for functions, which behave like this. Should I rename them all? (Seems like now all functions behave like that and to have long names for most of public fucntions may be weird)

PS: I have also changed a comment a bit: this functions will not just return from function call, it will return `false` in case, when active blob already exists

src/storage/core.rs
```diff
@@ -133,8 +151,84 @@ impl<K: Key> Storage<K> {
...
+    /// [`create_active_blob_async()`]: struct.Storage.html#method.create_active_blob_async
+    pub async fn create_active_blob(&self) -> Result<bool> {
+        self.inner.create_active_blob().await
+    }
+
+    /// Creates active blob
+    /// NOTICE! This function returns immediately, so you can't check result of operation. If you
+    /// want be sure about operation's result, use [`create_active_blob()`]
+    /// [`create_active_blob()`]: struct.Storage.html#method.create_active_blob
+    pub async fn create_active_blob_async(&self) {
```
idruzhitskiy, line 187: Although it's "async" indeed, I think it's better to change the name so it's easier to understand why it's different from non-`_async` variant. Something like `_non_blocking`.

src/storage/core.rs
```diff
@@ -446,29 +537,41 @@ impl<K: Key> Storage<K> {
...
             .store(safe.max_id().await.map_or(0, |i| i + 1), ORD);
         Ok(())
     }
 
+    async fn pop_active(blobs: &mut Vec<Blob>, config: &Config) -> Result<Box<Blob>> {
+        let mut active_blob = blobs
+            .pop()
+            .ok_or_else(|| {
+                let wd = config.work_dir();
+                error!("There are some blob files in the work dir: {:?}", wd);
```
idruzhitskiy, line 566: Why not single error message?

src/storage/observer_worker.rs
```diff
@@ -51,6 +58,57 @@ impl ObserverWorker {
...
+        } else {
+            true
+        }
+    }
+
+    async fn close_active_blob(&self, msg: Msg) -> Result<()> {
+        if self.predicate_wrapper(&msg.predicate).await {
+            let _lock = self.async_oplock.lock().await;
+            if self.predicate_wrapper(&msg.predicate).await {
+                self.inner.close_active_blob().await?;
```
idruzhitskiy, line 73: Lock logic is duplicated in all 4 msg processors, and content of the message is not used for any specific work. Probably it's better to move this logic into separate func (or into tick), and use `optype` inside this func to select `inner` action?

src/storage/observer.rs
```diff
@@ -4,39 +4,109 @@ use tokio::{
...
+            .await
+    }
+
+    async fn send_msg(&self, msg: Msg) {
         if let Some(sender) = &self.sender {
-            if let Err(e) = sender.send(Msg::CloseActiveBlob).await {
-                error!("observer cannot close active blob: task failed: {}", e);
+            if let Err(e) = sender.send(msg).await {
+                error!(
+                    "observer cannot force update active blob: task failed: {}",
```
idruzhitskiy, line 107: Error should be more generic and include msg type

src/storage/observer_worker.rs
Justarone: I moved it to tick and remembered, what I didn't like in this solution: now this function looks like it's written in LISP :D
It's not a problem for me, but firstly it looked weird)

Justarone: Resolve conversation if now it looks as you expected please

src/storage/core.rs
```diff
@@ -109,6 +109,24 @@ impl<K: Key> Storage<K> {
...
+    /// doesn't set active blob, which means that first write may take time..
+    ///
+    /// Storage works in directory provided to builder. If directory don't exist,
+    /// storage creates it, otherwise tries to init existing storage.
+    /// # Errors
+    /// Returns error in case of failures with IO operations or
+    /// if some of the required params are missed.
+    ///
+    /// [`init_noactive()`]: struct.Storage.html#method.init
+    pub async fn init_noactive(&mut self) -> Result<()> {
```
pyakushin, line 125: Just thought, maybe name it `init_lazy()`. Similar to rust vars lazy initialization

src/storage/core.rs
```diff
@@ -133,8 +151,84 @@ impl<K: Key> Storage<K> {
...
+    /// Fails if there are some errors during dump
+    /// [`close_active_blob_non_blocking()`]: struct.Storage.html#method.create_active_blob_async
+    pub async fn close_active_blob(&self) -> Result<bool> {
+        self.inner.close_active_blob().await
+    }
+
+    /// Dumps active blob
+    /// NOTICE! This function returns immediately, so you can't check result of operation. If you
+    /// want be sure about operation's result, use [`close_active_blob()`]
+    pub async fn nlose_active_blob_non_blocking(&self) {
```
pyakushin, line 205: Typo? `nlose`

src/storage/core.rs
```diff
@@ -133,8 +151,84 @@ impl<K: Key> Storage<K> {
...
+    /// [`create_active_blob_non_blocking()`]: struct.Storage.html#method.create_active_blob_async
+    pub async fn create_active_blob(&self) -> Result<bool> {
+        self.inner.create_active_blob().await
+    }
+
+    /// Creates active blob
+    /// NOTICE! This function returns immediately, so you can't check result of operation. If you
+    /// want be sure about operation's result, use [`create_active_blob()`]
+    /// [`create_active_blob()`]: struct.Storage.html#method.create_active_blob
+    pub async fn create_active_blob_non_blocking(&self) {
```
pyakushin, line 187: Maybe it's better to rename this kind of fns:
`create_active_blob` -> `create_active_blob_spawned` or `create_active_blob_trigger` or `create_active_blob_in_background`
`create_active_blob_non_blocking` -> `create_active_blob`
Because it's kind of misleading for async fn to be `blocking`.

src/storage/core.rs
pyakushin: I think it's better to use `Error::ActiveBlobExists` instead of bool. And maybe rename to `try_create_active_blob()`

src/storage/core.rs
```diff
@@ -621,6 +725,83 @@ impl Inner {
...
+            return Ok(false);
+        }
+        let mut safe = self.safe.write().await;
+        if let None = safe.active_blob {
+            let blob_opt = safe.blobs.write().await.pop().map(|b| b.boxed());
+            if let Some(blob) = blob_opt {
+                safe.active_blob = Some(blob);
+                Ok(true)
+            } else {
+                // NOTE: maybe it would be better panic in case of this operation with empty blob list?
```
pyakushin, line 739: If you think that it is highly unlikely for this case to happen, you can use expect or panic.
Either way, this case should be an Error::Unitialized

src/storage/core.rs
idruzhitskiy: It's a bit tricky, as `non_blocking` in this case refers to the function invocation and not to the underlying IO calls, and is a widespread term for this type of functions. Other terms can be misunderstood

ikopylov:
@pyakushin @idruzhitskiy @vovac12 please, take a look

PR #121, opened 2021-08-15
115-offloadable-bloom-filter

### LINKED ISSUE #115: Make bloom filter offloadable from memory
Bloom filter is stored only in RAM, so with the large number of BLOBs this leads to crash by Out Of Memory. We need to implement Bloom filter offloading and its work from index file on disk.

Related to https://github.com/qoollo/bob/issues/260

vovac12:
Closes #115


ikopylov:
@vovac12 please, fix merge conflicts

src/storage/core.rs
```diff
@@ -765,19 +785,36 @@ impl<K: Key> Storage<K> {
...
         let in_closed = inner
             .blobs
             .read()
             .await
             .iter()
-            .any(|blob| blob.check_filters(key.as_ref()) == true);
+            .map(|blob| blob.check_filters(key.as_ref()))
+            .collect::<FuturesUnordered<_>>()
+            .any(|value| value)
+            .await;
```
ikopylov, line 801: For performance reasons it is better to split this check into two sequential:
1. Check only those filters that are in memory. This is a quick check, and if it returns `true`, we don't need to perfrom checks on disk
2. If the previous step returned false, then run checks on disk

src/blob/index/core.rs
```diff
@@ -152,40 +180,43 @@ impl<FileIndex: FileIndexTrait> IndexStruct<FileIndex> {
...
 
-    fn deserialize_filters(buf: &[u8]) -> Result<(Bloom, RangeFilter)> {
+    fn deserialize_filters(buf: &[u8]) -> Result<(Bloom, RangeFilter, usize)> {
         let (range_size_buf, rest_buf) = buf.split_at(size_of::<u64>());
         let range_size = deserialize(&range_size_buf)?;
         let (range_buf, bloom_buf) = rest_buf.split_at(range_size);
         let bloom = Bloom::from_raw(bloom_buf)?;
         let range = RangeFilter::from_raw(range_buf)?;
-        Ok((bloom, range))
+        Ok((bloom, range, range_size + size_of::<u64>()))
```
ikopylov, line 208: I think it would be nice to add the `offset` field directly to the `Bloom` structure. By default it has a `None` value, but when the filter is dumped to disk or created from disk, the value in this field updated to be the actual offset. This approach allow to remove that additional element in the tuple

src/storage/core.rs
vovac12: Done

src/blob/index/core.rs
vovac12: Done

src/blob/index/core.rs
```diff
@@ -199,24 +212,24 @@ impl<FileIndex: FileIndexTrait> IndexStruct<FileIndex> {
         Ok((buf, bloom_offset))
     }
 
-    fn deserialize_filters(buf: &[u8]) -> Result<(Bloom, RangeFilter, usize)> {
+    fn deserialize_filters(buf: &[u8]) -> Result<(Bloom, RangeFilter)> {
         let (range_size_buf, rest_buf) = buf.split_at(size_of::<u64>());
         let range_size = deserialize(&range_size_buf)?;
         let (range_buf, bloom_buf) = rest_buf.split_at(range_size);
-        let bloom = Bloom::from_raw(bloom_buf)?;
+        let mut bloom = Bloom::from_raw(bloom_buf)?;
```
ikopylov, line 219: It seems reasonable for `from_raw` to take an additional optional argument with an offset

src/storage/core.rs
```diff
@@ -795,12 +795,24 @@ impl<K: Key> Storage<K> {
...
             .iter()
-            .map(|blob| blob.check_filters(key.as_ref()))
-            .collect::<FuturesUnordered<_>>()
-            .any(|value| value)
-            .await;
+            .filter(|blob| !blob.is_filter_offloaded())
+            .any(|blob| blob.check_filters_in_memory(key.as_ref()));
 
-        Some(in_active || in_closed)
+        if !(in_active || in_closed) {
```
ikopylov, line 801: I think the code will become more readable if you write it like this: 
```
let in_active = ...;
if (in_active)
    return Some(true);
let in_closed = ...;
if (in_closed)
    return Some(true);
let in_closed_offloaded = ...;
return Some(in_closed_offloaded)
```

src/storage/core.rs
```diff
@@ -795,12 +795,24 @@ impl<K: Key> Storage<K> {
...
+            let in_closed_offloaded = inner
+                .blobs
+                .read()
+                .await
+                .iter()
+                .filter(|blob| blob.is_filter_offloaded())
```
ikopylov, line 807: Is it possible that the bloom filter will be offloaded between the previous and current checks? If possible, the BLOB sets may diverge. In general, it would be safer to build a set of BLOBs with non-offloaded filters first, and then test the filters on BLOBs in that set and for blobs that are not in that set

src/blob/index/bloom.rs
```diff
@@ -137,34 +176,87 @@ impl Bloom {
         Ok(Self::from(save))
     }
 
-    pub fn add(&mut self, item: impl AsRef<[u8]>) {
+    pub fn add(&mut self, item: impl AsRef<[u8]>) -> Result<()> {
```
idruzhitskiy, line 179: I think both `add` and `contains_in_memory` have plenty in common (but `add` lacks 0 check), and should be unified. You can extract function, which accepts function `usize -> bool`, and use `all` in both cases.

src/blob/index/bloom.rs
```diff
@@ -137,34 +176,87 @@ impl Bloom {
...
+
+    pub async fn contains_in_file<P: BloomDataProvider>(
+        &self,
+        provider: &P,
+        item: impl AsRef<[u8]>,
+    ) -> Result<bool> {
         let mut hashers = self.hashers.clone();
-        let len = self.inner.len() as u64;
-        for h in hashers.iter_mut().map(|hasher| {
+        if self.bits_count == 0 {
```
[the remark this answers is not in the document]
vovac12: Done

src/blob/index/bloom.rs
vovac12: I have used another solution, please check it

src/storage/core.rs
vovac12: Done

src/storage/core.rs
vovac12: Done

src/blob/index/core.rs
vovac12: Done

src/blob/index/bloom.rs
idruzhitskiy: LGTM

src/blob/index/bloom.rs
```diff
@@ -137,34 +176,87 @@ impl Bloom {
...
-    pub fn add(&mut self, item: impl AsRef<[u8]>) {
+    pub fn add(&mut self, item: impl AsRef<[u8]>) -> Result<()> {
+        if let Some(inner) = &mut self.inner {
+            let mut hashers = self.hashers.clone();
+            let len = inner.len() as u64;
+            for h in hashers.iter_mut().map(|hasher| {
+                hasher.write(item.as_ref());
+                hasher.finish() % len
+            }) {
+                *inner
```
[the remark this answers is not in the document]
pyakushin: I'm against unsafe code) I don't think that it give such great performance increase to risk for code safety.

src/blob/index/bptree/core.rs
```diff
@@ -85,6 +85,18 @@ impl FileIndexTrait for BPTreeFileIndex {
         Ok(buf)
     }
 
+    async fn read_meta_at(&self, i: u64) -> Result<u8> {
+        trace!("load byte from meta");
+        if i >= self.header.meta_size as u64 {
+            return Err(anyhow::anyhow!("read meta out of range"));
+        }
+        let mut buf = vec![0; 1];
```
pyakushin, line 93: You can use just an array here

src/blob/index/core.rs
```diff
@@ -93,16 +94,56 @@ impl<FileIndex: FileIndexTrait> IndexStruct<FileIndex> {
...
+            self.bloom_filter.offload_from_memory();
+        }
+    }
+
+    pub(crate) async fn check_filters_key(&self, key: &[u8]) -> FilterResult {
         if !self.range_filter.contains(key) {
             FilterResult::NotContains
-        } else if self.params.bloom_is_on && !self.bloom_filter.contains(key) {
+        } else if self.params.bloom_is_on
+            && matches!(self.check_bloom_key(key).await, Ok(Some(false)))
```
pyakushin, line 107: It works fine. But we need to investigate errors somehow. And errors would be just dropped here. Maybe it makes sense to return `Result<FilterResult>` here

src/blob/index/core.rs
```diff
@@ -93,16 +94,56 @@ impl<FileIndex: FileIndexTrait> IndexStruct<FileIndex> {
...
         } else {
             FilterResult::NeedAdditionalCheck
         }
     }
 
+    pub(crate) fn check_filters_in_memory(&self, key: &[u8]) -> FilterResult {
+        if !self.range_filter.contains(key) {
+            FilterResult::NotContains
+        } else if self.params.bloom_is_on
+            && self.bloom_filter.contains_in_memory(key) == Some(false)
```
pyakushin, line 119: You can squash this condition with previous

src/blob/index/core.rs
```diff
@@ -93,16 +94,56 @@ impl<FileIndex: FileIndexTrait> IndexStruct<FileIndex> {
...
+            FilterResult::NeedAdditionalCheck
+        }
+    }
+
+    pub async fn check_bloom_key(&self, key: &[u8]) -> Result<Option<bool>> {
+        if self.params.bloom_is_on {
+            if let Some(result) = self.bloom_filter.contains_in_memory(key) {
+                Ok(Some(result))
+            } else {
+                Ok(Some(self.bloom_filter.contains_in_file(self, key).await?))
```
pyakushin, line 132: If understand correctly, you check file on every negative response from filter in memory. What if filter loaded and returned negative?

src/storage/core.rs
```diff
@@ -505,6 +505,26 @@ impl<K: Key> Storage<K> {
         }
     }
 
+    /// Returns memory allocated for bloom filter buffer
+    pub async fn filter_memory_allocated(&self) -> usize {
+        let safe = self.inner.safe.read().await;
+        let active = if let Some(ablob) = safe.active_blob.as_ref() {
```
pyakushin, line 511: It would be shorter with `map_or()`

pyakushin (changes requested):

src/blob/index/core.rs
vovac12: No, `contains_in_memory` returns `None` if the filter is offloaded and the file is checked only in this case

PR #126, opened 2021-09-08
Hierarchical filters

vovac12:


Cargo.toml
```diff
@@ -29,7 +33,8 @@ env_logger = "0.8"
 futures = "0.3"
 log = "0.4"
 rand = "0.8"
-rio = "0.9.4"
+#rio = "0.9.4"
+rio = { path = "../rio" }
```
pyakushin, line 37: Change back to crates.io version?

src/blob/core.rs
```diff
@@ -320,14 +316,45 @@ impl Blob {
...
 
-    pub(crate) fn check_bloom(&self, key: &[u8]) -> Option<bool> {
+    pub(crate) async fn check_bloom(&self, key: &[u8]) -> Result<Option<bool>> {
         trace!("check bloom filter");
-        self.index.check_bloom_key(key)
+        self.index.check_bloom_key(key).await
+    }
+
+    pub(crate) async fn check_bloom_simple(&self, key: &[u8]) -> Option<bool> {
+        match self.check_bloom(key).await {
```
pyakushin, line 325: Since you don't check error, you can do the same with: `self.check_bloom(key).await.ok()`

src/blob/core.rs
```diff
@@ -320,14 +316,45 @@ impl Blob {
...
     }
+
+    pub(crate) fn filter_memory_allocated(&self) -> usize {
+        self.index.bloom_memory_allocated()
+    }
+}
+
+#[async_trait::async_trait]
+impl BloomProvider for Blob {
+    type Key = [u8];
```
pyakushin, line 346: It's better to use `&[u8]` here

src/blob/index/benchmarks.rs
```diff
@@ -0,0 +1,226 @@
...
+            if let Some(v) = inmem.get_mut(&key) {
+                v.push(rh);
+            } else {
+                inmem.insert(key, vec![rh]);
+            }
+        });
+    inmem
+}
+
+fn generate_meta(meta_size: usize) -> Vec<u8> {
```
pyakushin, line 25: I think you don't need this method. And in fact you generate Vec not meta.

src/blob/index/bptree/core.rs
```diff
@@ -0,0 +1,359 @@
...
+        let root_node = Self::read_root(&file, metadata.tree_offset).await?;
+        Ok(Self {
+            file,
+            metadata,
+            header,
+            root_node,
+        })
+    }
+
+    fn file_size(&self) -> u64 {
```
pyakushin, line 68: I think this method is redundant, because it is not public so you already have access to file field

src/blob/index/bptree/core.rs
```diff
@@ -0,0 +1,359 @@
...
+        self.file.size()
+    }
+
+    fn records_count(&self) -> usize {
```
pyakushin, line 72: Same

src/bloom/hierarchical.rs
```diff
@@ -0,0 +1,335 @@
+use std::iter::FromIterator;
+
+use super::*;
+
+/// Container for types which is support bloom filtering
+#[derive(Debug)]
+pub struct HierarchicalBloom<Child>
+where
+    Child: BloomProvider + Send + Sync,
```
idruzhitskiy, line 9: If you can, you should try to avoid trait constraints on generic variables in struct definitions and only add them to impl clauses (as it's easier to add/remove new constraints later on)

src/bloom/hierarchical.rs
```diff
@@ -0,0 +1,335 @@
...
+
+impl Default for HierarchicalBloomInner {
+    fn default() -> Self {
+        Self::Vec(Default::default(), Default::default())
+    }
+}
+
+impl<Child> Default for HierarchicalBloom<Child>
+where
+    Child: BloomProvider + Send + Sync,
```
idruzhitskiy, line 32: You don't need `Send` and `Sync` here. There should be as little trait constraints as possible everywhere, as it helps with maintenance

src/bloom/hierarchical.rs
```diff
@@ -0,0 +1,335 @@
...
+        }
+    }
+}
+
+impl HierarchicalBloomInner {
+    #[async_recursion::async_recursion]
+    async fn check_filter<Child, Key>(&self, children: &[Child], item: &Key) -> Result<Option<bool>>
+    where
+        Child: BloomProvider<Key = Key> + Send + Sync,
+        Key: Send + Sync + AsRef<[u8]> + ?Sized,
```
idruzhitskiy, line 48: You probably should use global pearl's `Key` trait here

src/bloom/hierarchical.rs
```diff
@@ -0,0 +1,335 @@
...
+    {
+        match self {
+            HierarchicalBloomInner::Vec(filter, values) => {
+                if let Some(filter) = filter {
+                    if let Ok(true) = filter.contains_in_memory(item) {
+                        return Ok(Some(true));
+                    }
+                }
+                let mut falses = false;
```
idruzhitskiy, line 57: Looks like it always remains false

src/bloom/hierarchical.rs
```diff
@@ -0,0 +1,335 @@
...
+                    .get_mut(*child)
+                    .unwrap()
+                    .offload_buffer(needed_memory)
+                    .await
+            }
+        }
+    }
+
+    fn merge_filters(dest: &mut Option<Bloom>, source: &Option<Bloom>) {
+        match source {
```
idruzhitskiy, line 111: `source.zip(dest).map(...)`

src/bloom/hierarchical.rs
```diff
@@ -0,0 +1,335 @@
...
+    }
+
+    fn push_group(&mut self, value: Self) {
+        match self {
+            Self::Vec(filter, values) => {
+                match &value {
+                    Self::Vec(child_filter, _) => {
+                        Self::merge_filters(filter, child_filter);
+                    }
+                    _ => unreachable!(),
```
idruzhitskiy, line 132: I think you should mention `Item` variant explicitly everywhere, otherwise if any other variant will be added in the future no compiler warning will be issued

src/bloom/hierarchical.rs
```diff
@@ -0,0 +1,335 @@
...
+    fn push(&mut self, id: usize, filter: &Option<Bloom>, group_size: usize) {
+        match self {
+            Self::Vec(_, values) => {
+                if values.is_empty() {
+                    values.push(Default::default())
+                }
+                let last = values.last_mut().expect("Have values");
+                last.push_item(id, filter);
+                if last.len() >= group_size {
+                    drop(last);
```
idruzhitskiy, line 159: Is this manual call to `drop` necessary?

src/bloom/hierarchical.rs
```diff
@@ -0,0 +1,335 @@
...
+}
+
+impl HierarchicalBloomInner {
+    #[async_recursion::async_recursion]
+    async fn check_filter<Child, Key>(&self, children: &[Child], item: &Key) -> Result<Option<bool>>
+    where
+        Child: BloomProvider<Key = Key> + Send + Sync,
+        Key: Send + Sync + AsRef<[u8]> + ?Sized,
+    {
+        match self {
```
ikopylov, line 50: The logic seems to be wrong. Bloom filter can detect that key is definitely not presented (`false`). If it returns `true`, that means that key is presented with some probability. Thus, you must immediately return `false` when current level filter also returns `false`. If the current level filter returns `true`, then you need to go down the hierarchy

src/bloom/hierarchical.rs
```diff
@@ -0,0 +1,335 @@
...
+{
+    inner: HierarchicalBloomInner,
+    children: Vec<Child>,
+    group_size: usize,
+}
+
+#[derive(Debug)]
+enum HierarchicalBloomInner {
+    /// Inner filters
+    Vec(Option<Bloom>, Vec<HierarchicalBloomInner>),
```
ikopylov, line 19: Why bloom is declared as `Option<>` here?

src/bloom/hierarchical.rs
```diff
@@ -0,0 +1,335 @@
...
+        self.children.push(item);
+    }
+
+    /// Returns children elements as Vec
+    pub fn into_vec(self) -> Vec<Child> {
+        self.children
+    }
+
+    /// Returns mutable reference to inner container. Use carefully
+    pub fn children_mut(&mut self) -> &mut Vec<Child> {
```
ikopylov, line 274: Exposing inner collection looks unsafe. If it is changed externally, there will be an out of sync between `collection` and `inner`.

src/bloom/hierarchical.rs
```diff
@@ -0,0 +1,335 @@
...
+        match self {
+            Self::Vec(filter, values) => {
+                values.push(Self::Item(id));
+                Self::merge_filters(filter, child_filter)
+            }
+            _ => unreachable!(),
+        }
+    }
+
+    fn push(&mut self, id: usize, filter: &Option<Bloom>, group_size: usize) {
```
ikopylov, line 150: Please add comments describing how this method works

src/bloom/hierarchical.rs
```diff
@@ -0,0 +1,335 @@
...
+
+    fn add_to_intermediate_filters(&mut self, id: usize, item: &[u8]) -> Result<bool> {
+        match self {
+            Self::Vec(filter, values) => {
+                if let Some(filter) = filter {
+                    let mut should_add = false;
+                    for value in values.iter_mut() {
+                        should_add = should_add || value.add_to_intermediate_filters(id, item)?;
+                        if should_add {
+                            filter.add(item)?;
```
ikopylov, line 188: If you have a guarantee that only one item has specified `id` (ids are unique), then you should stop iterating over `values` here

src/bloom/hierarchical.rs
```diff
@@ -0,0 +1,335 @@
...
+                    .await
+            }
+        }
+    }
+
+    fn merge_filters(dest: &mut Option<Bloom>, source: &Option<Bloom>) {
+        match source {
+            Some(source_filter) => {
+                let res = match dest {
+                    Some(dest_filter) => dest_filter.checked_add_assign(source_filter),
```
ikopylov, line 114: `checked_add_assign` may fail if the lengths are different. You should handle this in some way. The best way is to prevent merging (two separate filters will be kept). The easy way is to panic

src/bloom/hierarchical.rs
```diff
@@ -0,0 +1,335 @@
...
+    fn default() -> Self {
+        Self {
+            inner: Default::default(),
+            children: Default::default(),
+            group_size: 8,
+        }
+    }
+}
+
+impl HierarchicalBloomInner {
```
ikopylov, line 43: There should also be a method, that allows to get the list of BLOBs that may contain the key. It is not enough just to say, that some BLOB has specified key. With hierarchical filters we can reduce the number of BLOBs to scan by rejecting the BLOB groups that definitely do not contain the key

ikopylov (changes requested):
Changes are required

src/bloom/hierarchical.rs
vovac12: Done in new version

src/bloom/hierarchical.rs
```diff
@@ -0,0 +1,610 @@
...
+        let ret = *self;
+        self.0 += 1;
+        ret
+    }
+}
+
+/// Container for types which is support bloom filtering
+pub struct HierarchicalBloom<Child> {
+    inner: BTreeMap<InnerId, HierarchicalBloomInner>,
+    children: BTreeMap<ChildId, Leaf<Child>>,
```
ikopylov, line 32: Is it possible to declare `children` as a vector, so that `ChildId` be an index inside that vector?

pyakushin (changes requested):
Check tests

src/bloom/hierarchical.rs
vovac12: Yeah, I've done it

ikopylov:
Please, resolve conflicts

ikopylov:
Please, add in comments a short description of how most critical methods work

pyakushin (approved):

idruzhitskiy (approved):
A couple of style questions, otherwise LGTM

ikopylov (approved):
Please, add comments about work logic to the following methods in `hierarchical.rs`: `offload_buffer`, `push`, `add_child` and `Iterator::next`. It is hard to understand what is going on there. 

Otherwise, LGTM

ikopylov:
Issue for test failures: https://github.com/qoollo/pearl/issues/144

PR #129, opened 2021-10-01
add blob version validation

### LINKED ISSUE #120: Add validation for blob magic_byte and version

pyakushin:
Closes #120 
Blob magic byte check already implemented

### DIFF
--- a/src/blob/header.rs
+++ b/src/blob/header.rs
@@ -6,18 +6,18 @@ use crate::{blob::File, error::ValidationParam, Error};
 
 use super::FileName;
 
-crate const BLOB_VERSION: u32 = 1;
-crate const BLOB_MAGIC_BYTE: u64 = 0xdeaf_abcd;
+pub(crate) const BLOB_VERSION: u32 = 1;
+pub(crate) const BLOB_MAGIC_BYTE: u64 = 0xdeaf_abcd;
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
-crate struct Header {
+pub(crate) struct Header {
     magic_byte: u64,
     version: u32,
     flags: u64,
 }
 
 impl Header {
-    crate const fn new() -> Self {
+    pub(crate) const fn new() -> Self {
         Self {
             magic_byte: BLOB_MAGIC_BYTE,
             version: BLOB_VERSION,
@@ -25,7 +25,7 @@ impl Header {
         }
     }
 
-    crate async fn from_file(name: &FileName, ioring: Option<Rio>) -> Result<Self> {
+    pub(crate) async fn from_file(name: &FileName, ioring: Option<Rio>) -> Result<Self> {
         let file = File::open(name.to_path(), ioring)
             .await
             .with_context(|| format!("failed to open blob file: {}", name))?;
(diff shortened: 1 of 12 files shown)


src/blob/index/bptree/core.rs
```diff
@@ -315,10 +321,12 @@ impl BPTreeFileIndex {
     async fn validate_header(&self, buf: &mut Vec<u8>) -> Result<()> {
         self.validate()?;
         if self.header.version() != HEADER_VERSION {
```
vovac12, line 323: Header version already checked in `self.validate()`

src/blob/index/simple.rs
```diff
@@ -240,10 +246,12 @@ impl SimpleFileIndex {
     async fn validate_header(&self, buf: &mut Vec<u8>) -> Result<()> {
         self.validate()?;
         if !Self::hash_valid(&self.header, buf)? {
-            return Err(Error::validation("header hash mismatch").into());
+            let param = ValidationParam::IndexChecksum;
+            return Err(Error::validation(param, "header hash mismatch").into());
         }
         if self.header.version() != HEADER_VERSION {
```
vovac12, line 252: Same

src/error.rs
```diff
@@ -115,8 +116,49 @@ pub enum Kind {
...
+        cause: String,
+    },
+
     /// Other error
     Other,
 }
+
+/// Variants of validation errors.
+#[derive(Debug, Clone, PartialEq)]
+pub enum ValidationParam {
```
vovac12, line 133: Maybe `ValidationErrorKind` is more descriptive name? And also `param` -> `kind` in field name

src/lib.rs
```diff
@@ -1,3 +1,4 @@
+#![feature(crate_visibility_modifier)]
```
vovac12, line 1: Is it really necessary? Pearl will not compile on stable rust with it.

vovac12 (changes requested):

src/storage/core.rs
```diff
@@ -573,28 +583,34 @@ impl<K: Key> Storage<K> {
     }
 
     fn should_save_corrupted_blob(error: &anyhow::Error) -> bool {
+        debug!("decide wether to save corrupted blobs: {:#}", error);
         if let Some(error) = error.downcast_ref::<Error>() {
             return match error.kind() {
-                ErrorKind::Bincode(_) | ErrorKind::Validation(_) => true,
+                ErrorKind::Bincode(_) => true,
+                ErrorKind::Validation { param, cause: _ }
+                    if param != &ValidationParam::BlobVersion =>
```
idruzhitskiy, line 591: IMO It's better to use inner match with two branches - BlobVersion and default

idruzhitskiy (approved):

src/lib.rs
pyakushin: Removed

vovac12 (approved):

src/blob/header.rs
```diff
@@ -0,0 +1,58 @@
...
+            .await
+            .with_context(|| format!("failed to read from file: {}", name))?;
+        let header: Self = deserialize(&buf)
+            .with_context(|| format!("failed to deserialize header from file: {}", name))?;
+        header.validate().context("header validation failed")?;
+        Ok(header)
+    }
+
+    fn validate(&self) -> Result<()> {
+        if self.version != BLOB_VERSION {
```
ikopylov, line 44: It is necessary to check `version` after `magic_byte`, because the version mismatch can be either due to a corrupted header or due to the fact that the BLOB was not updated. `magic_byte` check in the first place can resolve this situation

PR #131, opened 2021-10-01
Create corrupted dir only when needed

### LINKED ISSUE #130: Create `corrupted` directory only when corrupted BLOB is found
Currently `corrupted` directory is created at startup for every partition

vovac12:
Closes #130


ikopylov:
LGTM

pyakushin (approved):

PR #140, opened 2021-11-11
Add key type variable

idruzhitskiy:


pyakushin (approved):

src/blob/core.rs
```diff
@@ -13,15 +13,16 @@ pub(crate) const BLOB_INDEX_FILE_EXTENSION: &str = "index";
...
 /// [`Blob`]: struct.Blob.html
 #[derive(Debug)]
-pub struct Blob {
+pub struct Blob<K> {
     header: Header,
     index: Index,
     name: FileName,
     file: File,
     current_offset: Arc<Mutex<u64>>,
+    key_type: PhantomData<K>,
```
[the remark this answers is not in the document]
idruzhitskiy: Done

src/blob/index/range.rs
```diff
@@ -58,3 +44,178 @@ impl RangeFilter {
         bincode::serialize(&self).map_err(|e| e.into())
     }
 }
+
+const STRUCT_NAME: &str = "RangeFilter";
```
vovac12, line 48: Why custom (De)serialization implementation is needed

vovac12 (approved):

src/blob/index/range.rs
idruzhitskiy: Changed to usage of `(de)serialize_with` attribute for key fields

src/blob/index/bptree/node.rs
```diff
@@ -38,13 +38,15 @@ impl Node {
...
+        let key_size = K::LEN as usize;
         let mut l = 0i32;
-        let mut r: i32 = (buf.len() / key.len() - 1) as i32;
+        let mut r: i32 = (buf.len() / key_size - 1) as i32;
         while l <= r {
             let m = (l + r) / 2;
-            let offset = m as usize * key.len();
-            match key.cmp(&buf[offset..(offset + key.len())]) {
+            let offset = m as usize * key_size;
+            let k = buf[offset..(offset + key_size)].to_vec().into(); // TODO Allow keys to be based on slices
```
ikopylov, line 48: :+1: It would be great to avoid copying into a vector

src/blob/index/bptree/core.rs
```diff
@@ -148,8 +152,8 @@ impl FileIndexTrait for BPTreeFileIndex {
     }
 }
 
-impl BPTreeFileIndex {
-    async fn find_leaf_node(&self, key: &[u8], mut offset: u64, buf: &mut [u8]) -> Result<u64> {
+impl<K: Key> BPTreeFileIndex<K> {
```
ikopylov, line 155: `read_header_buf` in this file performs key comparison. It should be updated to use a method from `Key` trait

src/blob/index/bptree/node.rs
```diff
@@ -53,10 +55,10 @@ impl Node {
         Err(l as usize)
     }
 
-    pub(super) fn key_offset_serialized(buf: &[u8], key: &[u8]) -> Result<u64> {
+    pub(super) fn key_offset_serialized<K: Key>(buf: &[u8], key: &K) -> Result<u64> {
```
ikopylov, line 58: Please, take a look at `key_offset` method. It uses `binary_search_by` on keys, so it needs to be updated, but additionally it is marked with `#[allow(dead_code)]` so the method should be inspected. There are two ways of resolution: update method or remove it.

src/blob/index/simple.rs
```diff
@@ -101,8 +101,8 @@ impl FileIndexTrait for SimpleFileIndex {
             .map(|headers| (headers, self.header.records_count))
     }
 
-    async fn get_any(&self, key: &[u8]) -> Result<Option<RecordHeader>> {
-        Self::binary_search(&self.file, key, &self.header)
+    async fn get_any(&self, key: &K) -> Result<Option<RecordHeader>> {
+        Self::binary_search(&self.file, &key.to_vec(), &self.header)
```
ikopylov, line 105: `binary_search` definitely contains key comparison logic, so I guess it needs to be updated too

ikopylov:
I think we need to increment index file version, because after this update the order of record inside will be changed

ikopylov (changes requested):
Changes are required

src/blob/index/bptree/node.rs
idruzhitskiy: Yet I think it should be implemented in another ticker, as it will require lifetime to be added to `Key` trait

src/blob/index/bptree/node.rs
idruzhitskiy: `key_offset` method is used only in tests, but still I've updated it with simple conversions

src/blob/index/simple.rs
idruzhitskiy: Added key trait usage to this binary_search

src/blob/index/bptree/core.rs
idruzhitskiy: Added key trait usage to `read_header_buf` (with vec copy for now)

src/blob/index/bptree/node.rs
ikopylov: Have you measured the performance impact of this data copying? If it is large, we will need to implement support for slices in 2.0 milestone

pyakushin (approved):
LGTM

src/blob/index/bptree/node.rs
idruzhitskiy: It's about 10 microseconds more on index search, according to benchmarks. While working with file indexes it's not noticeable at all.

ikopylov: The relative perf change is more interesting in this case

idruzhitskiy: ~10% more per call to search (from ~70us to ~80us)

ikopylov: Ok, thanks! This regression is not too large, so we can fix it in a separate task

pyakushin:
@ikopylov please review recent changes and approve if everything is ok.

src/blob/index/benchmarks.rs
```diff
@@ -32,14 +32,14 @@ impl Default for KeyType {
...
+    let mut inmem = InMemoryIndex::<KeyType>::new();
     (0..records_amount as u32)
         .map(key_mapper)
         .map(|i| serialize(&i).expect("can't serialize"))
-        .for_each(|mut key| {
-            key.resize(KeyType::LEN as usize, 0);
-            let rh = RecordHeader::new(key.clone(), 1, 1, 1);
+        .for_each(|key| {
+            let key: KeyType = key.into();
+            let rh = RecordHeader::new(key.clone().to_vec(), 1, 1, 1);
```
ikopylov, line 42: Do we need to `clone` the key if it is immediately converted to a vector?

src/blob/index/bptree/serializer.rs
```diff
@@ -84,7 +84,7 @@ impl<'a> HeaderStage<'a> {
         let mut min_o = offset;
         for (k, v) in btree.iter() {
             if remainder < record_header_size {
-                leaf_nodes_compressed.push((min_k, min_o));
+                leaf_nodes_compressed.push((min_k.clone().to_vec(), min_o));
```
ikopylov, line 87: Same question

src/blob/index/bptree/serializer.rs
```diff
@@ -93,7 +93,7 @@ impl<'a> HeaderStage<'a> {
             offset += delta_size;
             remainder = remainder.saturating_sub(delta_size);
         }
-        leaf_nodes_compressed.push((min_k, min_o));
+        leaf_nodes_compressed.push((min_k.clone().to_vec(), min_o));
```
ikopylov, line 96: And here

src/blob/index/benchmarks.rs
idruzhitskiy: This code left from the experiments with consuming `to_vec`. Fixed in all cases

ikopylov (approved):
LGTM

Issue #144, opened 2021-12-08
Tests failure due to asynchronous index dumping


ikopylov:
Several tests (`test_multithread_read_write`, `test_index_from_blob`) fail, because they expect the index file to appear on disk. But after recent changes, index dumping occurs non-deterministically in the background

ikopylov:
PR with background index dumping: https://github.com/qoollo/pearl/pull/137

PR #146, opened 2021-12-13
144 test file wait

idruzhitskiy:


idruzhitskiy:
closes #144

tests/common.rs
```diff
@@ -125,8 +125,18 @@ pub async fn clean(storage: Storage<KeyTest>, path: impl AsRef<Path>) -> Result<
     fs::remove_dir_all(path).map_err(Into::into)
 }
 
-pub async fn close_storage(storage: Storage<KeyTest>) -> Result<()> {
-    std::thread::sleep(std::time::Duration::from_millis(100));
+const MAX_WAIT_CYCLES: usize = 10;
+const WAIT_DELAY: std::time::Duration = std::time::Duration::from_millis(20);
+
+pub async fn close_storage(storage: Storage<KeyTest>, expected_files: &[&PathBuf]) -> Result<()> {
```
ikopylov, line 131: You can generalize this approach: create method that takes a predicate as an argument and wait in loop for it to return `true`. In this form, the method can be used in many places by passing various predicates.

idruzhitskiy: I've added separate fn for this (but it's still used only here, since other tests doesn't have obvious predicates to use)

ikopylov:
LGTM

PR #147, opened 2022-01-06
Fix bug with last record header retrieval from in memory index

Justarone:
Records are stored in reversed order only on disk, in memory they are in straight order, so the last header should be retrieved

### DIFF
--- a/src/blob/index/core.rs
+++ b/src/blob/index/core.rs
@@ -322,7 +322,7 @@ where
         match &self.inner {
             State::InMemory(headers) => {
                 debug!("index get any in memory headers: {}", headers.len());
-                Ok(headers.get(key).and_then(|h| h.first()).cloned())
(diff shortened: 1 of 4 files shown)


src/blob/index/core.rs
```diff
@@ -322,7 +322,7 @@ where
         match &self.inner {
             State::InMemory(headers) => {
                 debug!("index get any in memory headers: {}", headers.len());
-                Ok(headers.get(key).and_then(|h| h.first()).cloned())
+                Ok(headers.get(key).and_then(|h| h.last()).cloned())
```
ikopylov, line 325: Please, add a comment on why the `last` is used here. This should help not to forget about it in future

ikopylov:
Thanks! Please, add a line to CHANGELOG.md and check failing tests

ikopylov:
LGTM

PR #148, opened 2022-01-21
Move brt util functions to pearl

vovac12:


src/record.rs
```diff
@@ -2,24 +2,36 @@ use crate::{error::ValidationErrorKind, prelude::*};
...
-    meta_size: u64,
-    data_size: u64,
+    pub meta_size: u64,
+    pub data_size: u64,
     flags: u8,
     blob_offset: u64,
     created: u64,
-    data_checksum: u32,
-    header_checksum: u32,
+    pub data_checksum: u32,
```
idruzhitskiy, line 33: Do we need this to be public?

idruzhitskiy: Also other fields should be made pub(crate)?

src/lib.rs
```diff
@@ -71,6 +71,9 @@ mod storage;
 pub mod filter;
 pub use filter::{Bloom, BloomDataProvider, BloomProvider, Config as BloomConfig, FilterResult};
 
+/// tools to interact with pearl structures
+pub mod tools;
```
idruzhitskiy, line 75: I think you can name this module more specific, like `recovery` or `reconstruction`

vovac12: But this is also `migration`

idruzhitskiy: Yes, you're right... Then I guess we have to use generic name

src/blob/index/header.rs
```diff
@@ -71,6 +73,14 @@ impl IndexHeader {
     pub(crate) fn from_raw(buf: &[u8]) -> bincode::Result<Self> {
         bincode::deserialize(buf)
     }
+
+    pub(crate) fn validate_without_version(&self) -> Result<()> {
+        if !self.is_written() {
+            let param = ValidationErrorKind::RecordMagicByte;
```
ikopylov, line 79: Validation error does not match detected problem

src/blob/header.rs
```diff
@@ -9,10 +9,10 @@ use super::FileName;
 pub(crate) const BLOB_VERSION: u32 = 1;
 pub(crate) const BLOB_MAGIC_BYTE: u64 = 0xdeaf_abcd;
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub(crate) struct Header {
-    magic_byte: u64,
-    version: u32,
+    pub magic_byte: u64,
+    pub version: u32,
```
ikopylov, line 15: What is the reason to make these fields public? And why only these 2 fields without `flags`?

src/tools/validation.rs
```diff
@@ -0,0 +1,48 @@
...
+    let header = read_index_header(path)?;
+    let headers = match header.version() {
+        2 => block_on(async {
+            let index =
+                <SimpleFileIndex as FileIndexTrait<K>>::from_file(FileName::from_path(path)?, None)
+                    .await?;
+            let res = index.get_records_headers().await?;
+            AnyResult::<_>::Ok(res.0)
+        })?,
+        3..=4 => block_on(async {
```
ikopylov, line 25: We will definitely forget to update this place if the version changes. I suggest adding a compile-time assertion that checks for `HEADER_VERSION` and generates a compile-time error if it differs from what is expected here. I found `static_assertion` crate that can do this check

src/tools/migration.rs
```diff
@@ -0,0 +1,35 @@
...
+    input: &Path,
+    output: &Path,
+    validate_every: usize,
+    target_version: u32,
+) -> AnyResult<()> {
+    process_blob_with(
+        &input,
+        &output,
+        validate_every,
+        |record, version| record.migrate(version, target_version),
```
ikopylov, line 31: We need to update `blob` version in header too. There was an error in `brt` that was fixed here: https://github.com/qoollo/bob/pull/449. This fix should be applied here

src/lib.rs
ikopylov: Maybe `blob_tools` or `file_tools`

src/record.rs
```diff
@@ -245,4 +226,47 @@ impl Header {
     fn crc32(&self) -> bincode::Result<u32> {
         self.to_raw().map(|raw| CRC32C.checksum(&raw))
     }
+
+    pub(crate) fn with_reversed_key_bytes(mut self) -> bincode::Result<Self> {
```
ikopylov, line 230: This is not a common method, it is needed for migrations only, thus I think it is better to place it in `migration.rs`

src/tools/blob_writer.rs
```diff
@@ -0,0 +1,110 @@
...
+        file.seek(SeekFrom::Start(0))?;
+        let mut reader = BlobReader::from_file(file)?;
+        let header = reader.read_header()?;
+        if header != *written_header {
+            return Err(Error::blob_header_validation_error(
+                "validation of written blob header failed",
+            )
+            .into());
+        }
+        self.file.seek(SeekFrom::Start(current_position))?;
```
ikopylov, line 75: As I see it, you are cloning the file descriptor above and reading from that clonned file. Probably, this operation on original file descriptor is not needed

src/tools/validation.rs
ikopylov: In fact, I think, that we don't need the facilities to check all possible index file versions. Because index can simply be regenerated if something wrong, then we just need to check for validity on the latest version. If index from some previous version, then it is error and it should be regenerated

src/tools/migration.rs
```diff
@@ -0,0 +1,52 @@
...
+        |header| header.migrate(target_version),
```
ikopylov, line 32: Perhaps it's better to have the same signature for `record.migrate` and `header.migrate` (pass source version as the first parameter and target version as the second one)?

src/tools/validation.rs
```diff
@@ -0,0 +1,47 @@
...
+/// Validate index file
+pub fn validate_index<K>(path: &Path) -> AnyResult<()>
+where
+    for<'a> K: Key<'a> + 'static,
+{
+    static_assertions::const_assert_eq!(blob::index::HEADER_VERSION, 5);
+    let header = read_index_header(path)?;
+    let blob_size = path.metadata()?.len();
+    let headers = match header.version() {
+        2..=4 => return Err(Error::index_header_validation_error("too low header version").into()),
```
ikopylov, line 23: I suggest to extend error message: `Index version is outdated. Passed version: {}, Current version: {}`

src/tools/validation.rs
```diff
@@ -0,0 +1,47 @@
...
+
+/// Validate index file
+pub fn validate_index<K>(path: &Path) -> AnyResult<()>
+where
+    for<'a> K: Key<'a> + 'static,
+{
+    static_assertions::const_assert_eq!(blob::index::HEADER_VERSION, 5);
+    let header = read_index_header(path)?;
+    let blob_size = path.metadata()?.len();
+    let headers = match header.version() {
```
ikopylov, line 22: It looks like transforming this into `if...else` would simplify the code. Something like this:
```
if header.version() < blob::index::HEADER_VERSION {
    ...
}
else if header.version() == blob::index::HEADER_VERSION {
    ...
} 
else {
    ...
}
```

src/record.rs
vovac12: But I will need to make public `key` field and `update_checksum` method which is not good IMO

src/error.rs
```diff
@@ -165,6 +165,8 @@ pub enum ValidationErrorKind {
     RecordMagicByte,
     /// Index blob size
     IndexBlobSize,
+    /// Index is not written (index header corrupted)
+    IndexNotWritten,
```
ikopylov, line 169: There is `IndexIsWritten` error above. It seems to me that it was mistakenly used for cases where the index is not actually written.  Please, check all places where `IndexIsWritten` is generated. If you find a place where its usage is incorrect, replace it with the new `IndexNotWritten`

vovac12: Done

src/record.rs
ikopylov: Ok, I agree

src/blob/header.rs
ikopylov: @vovac12  Please, answer this question

vovac12: Made all fields `pub(crate)`

src/record.rs
vovac12: Done

ikopylov (approved):
LGTM

PR #151, opened 2022-02-21
Functions to get information about blobs and indexes

vovac12:


src/blob/index/core.rs
```diff
@@ -370,6 +370,7 @@ pub(crate) trait FileIndexTrait<K>: Sized + Send + Sync {
     ) -> Result<Self>;
     fn file_size(&self) -> u64;
     fn records_count(&self) -> usize;
+    fn blob_size(&self) -> u64;
```
ikopylov, line 373: `HEADER_VERSION` in this file should be incremented, because index file format has changed

src/tools/generic_key.rs
```diff
@@ -0,0 +1,56 @@
+use super::prelude::*;
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct GenericKey<const N: usize>([u8; N]);
```
ikopylov, line 4: We now have `ArrayKey<>` struct, so this `GenericKey<>` should be removed and `ArrayKey<>` should be used instead

src/tools/collectors.rs
```diff
@@ -0,0 +1,144 @@
...
+#[derive(Debug)]
+pub struct BlobSummaryCollector {
+    header: BlobHeader,
+    records: usize,
+    keys: BTreeSet<Vec<u8>>,
+}
+
+impl BlobSummaryCollector {
+    /// Collect data about blob by path
+    pub fn from_path(path: &Path) -> Result<Self> {
```
ikopylov, line 15: We should be able to collect both short and full information about BLOB. Short information is just an information from BLOB header, full is information obtained by BLOB scanning. I suggest to store a path to the file inside struct and add 2 functions: first loads data from header, second - loads data by scanning the whole BLOB file. `from_path` should receive a `bool` argument, that allows full loading when it is true.

src/tools/collectors.rs
```diff
@@ -0,0 +1,144 @@
+use std::collections::BTreeSet;
+
+use super::prelude::*;
+
+/// Collect data about blob
+#[derive(Debug)]
+pub struct BlobSummaryCollector {
+    header: BlobHeader,
+    records: usize,
+    keys: BTreeSet<Vec<u8>>,
```
ikopylov, line 10: Now we have records with `deleted` flag, so we should count the number of such records and also should store all unique deleted keys

src/tools/utils.rs
```diff
@@ -115,3 +119,53 @@ where
     );
     Ok(())
 }
+
+/// block on function
+pub fn block_on<T, F: Future<Output = T>>(f: F) -> Result<T> {
+    Ok(tokio::runtime::Runtime::new()?.block_on(f))
```
ikopylov, line 125: I think, it should try to use the current runtime if available. If not - then create a new one

src/tools/utils.rs
```diff
@@ -115,3 +119,53 @@ where
     );
     Ok(())
 }
+
+/// block on function
+pub fn block_on<T, F: Future<Output = T>>(f: F) -> Result<T> {
```
ikopylov, line 124: I think, this function should not be a public

src/tools/utils.rs
```diff
@@ -115,3 +119,53 @@ where
...
+    Ok(tokio::runtime::Runtime::new()?.block_on(f))
+}
+
+fn index_from_file<K>(
```
ikopylov, line 128: This function should be async

src/tools/utils.rs
```diff
@@ -115,3 +119,53 @@ where
...
+    }?;
+    let headers = headers
+        .into_iter()
+        .map(|(key, value)| (key.to_vec(), value))
+        .collect();
+    Ok(headers)
+}
+
+/// Read index file
+pub fn read_index(path: &Path) -> AnyResult<BTreeMap<Vec<u8>, Vec<record::Header>>> {
```
ikopylov, line 154: This function should have 2 implementation:
1. Async with all logic
2. Sync, that calls async one with the help of `block_on` defined above

src/tools/utils.rs
```diff
@@ -115,3 +119,53 @@ where
...
+
+async fn index_from_file<K>(
+    header: &IndexHeader,
+    path: &Path,
+) -> AnyResult<BTreeMap<Vec<u8>, Vec<record::Header>>>
+where
+    for<'a> K: Key<'a> + 'static
+{
+    let headers = match header.version() {
+        5 => block_on(async {
```
ikopylov, line 136: Current `HEADER_VERSION` constant should be used here.

Also, the idea behind making this function `async` was to remove this `block_on` call from its body

src/tools/collectors.rs
```diff
@@ -0,0 +1,171 @@
...
+            if record.header().is_deleted() {
+                collector.add_deleted_record(record);
+            } else {
+                collector.add_record(record);
+            }
+        }
+        Ok(collector)
+    }
+
+    fn load_short(path: &Path) -> Result<Self> {
```
ikopylov, line 40: `load_header_only`

src/tools/collectors.rs
```diff
@@ -0,0 +1,171 @@
+use std::collections::BTreeSet;
+
+use super::prelude::*;
+
+/// Collect data about blob
+#[derive(Debug)]
+pub struct BlobSummaryCollector {
+    header: BlobHeader,
+    records: usize,
+    deleted_records: usize,
```
ikopylov, line 10: You should expose new values through access functions in the same way the other values exposed below

src/tools/utils.rs
```diff
@@ -115,3 +119,53 @@ where
...
+    }?;
+    let headers = headers
+        .into_iter()
+        .map(|(key, value)| (key.to_vec(), value))
+        .collect();
+    Ok(headers)
+}
+
+/// Read index file
+pub async fn read_index(path: &Path) -> AnyResult<BTreeMap<Vec<u8>, Vec<record::Header>>> {
```
ikopylov, line 154: sync version of this function should also be added

src/tools/utils.rs
Boneyan: It is also used in tools/validation/validate_index()

src/blob/index/bptree/core.rs
```diff
@@ -14,7 +15,10 @@ use super::prelude::*;
 pub(super) const BLOCK_SIZE: usize = 4096;
 
 #[derive(Debug, Clone)]
-pub(crate) struct BPTreeFileIndex<K> {
+pub(crate) struct BPTreeFileIndex<K>
+where
```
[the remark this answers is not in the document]
Boneyan: No

src/blob/index/header.rs
```diff
@@ -39,13 +42,15 @@ impl IndexHeader {
         meta_size: usize,
         hash: Vec<u8>,
         blob_size: u64,
+        key_size: u16,
```
ikopylov, line 45: Different order of parameters in `with_hash` and in `new`. This increases the chance of error. Place `key_size` before `blob_size` as it is in `new` function

src/blob/index/header.rs
```diff
@@ -39,13 +42,15 @@ impl IndexHeader {
...
     ) -> Self {
         Self {
             records_count,
             record_header_size,
             meta_size,
             hash,
             blob_size,
+            key_size,
```
ikopylov, line 53: It also would good to change the order here

src/tools/utils.rs
```diff
@@ -115,3 +119,58 @@ where
     );
     Ok(())
 }
+
+/// block on function
+pub(super) fn block_on<T, F: Future<Output = T>>(f: F) -> Result<T> {
+    Ok(tokio::runtime::Handle::current().block_on(f))
```
ikopylov, line 125: This implementation will panic if there is no runtime. You should check, whether the runtime is available. If it is then use `tokio::runtime::Handle::current()`, if it is not then create a new runtime

src/tools/utils.rs
```diff
@@ -115,3 +119,58 @@ where
...
+
+async fn index_from_file<K>(
+    header: &IndexHeader,
+    path: &Path,
+) -> AnyResult<BTreeMap<Vec<u8>, Vec<record::Header>>>
+where
+    for<'a> K: Key<'a> + 'static
+{
+    let headers = match header.version() {
+        6 => {
```
ikopylov, line 136: Use `HEADER_VERSION` constant here from src/blob/index/core.rs

src/tools/utils.rs
ikopylov: Ok. Didn't notice that

ikopylov (approved):
LGTM

PR #154, opened 2022-02-28
Hierarchical range filters

vovac12:


src/blob/index/core.rs
```diff
@@ -70,13 +69,16 @@ pub(crate) enum State<FileIndex, K> {
 impl<FileIndex: FileIndexTrait<K>, K: Key> IndexStruct<FileIndex, K> {
     pub(crate) fn new(name: FileName, ioring: Option<Rio>, config: IndexConfig) -> Self {
         let params = IndexParams::new(config.bloom_config.is_some(), config.recreate_index_file);
-        let filter = config.bloom_config.map(Bloom::new).unwrap_or_default();
+        let filter = if params.bloom_is_on {
```
ikopylov, line 72: It's better to name it `bloom_filter`

src/storage/core.rs
```diff
@@ -390,7 +393,10 @@ impl<K: Key + 'static> Storage<K> {
         let possible_blobs = blobs
             .iter_possible_childs_rev(key)
             .map(|(id, blob)| async move {
-                if matches!(blob.data.check_filters(key).await, Ok(true)) {
+                if matches!(
+                    blob.data.check_filter(key).await,
+                    FilterResult::NeedAdditionalCheck
```
ikopylov, line 398: I'd prefer to inverse this check: compare for inequality with `FilterResult::NoContains`. It would be safer if `FilterResult` will be extended some day

ikopylov (approved):
LGTM

PR #160, opened 2022-04-05
should save corrupted blob on 'unexpected eof' error

Boneyan:
Corrupted blob now would be saved in case of 'unexpected eof' error

### DIFF
--- a/src/storage/core.rs
+++ b/src/storage/core.rs
@@ -3,7 +3,6 @@ use crate::error::ValidationErrorKind;
 use super::prelude::*;
 use futures::stream::FuturesOrdered;
 use tokio::fs::{create_dir, create_dir_all};
-use std::io::ErrorKind as IOErrorKind;
 
 const BLOB_FILE_EXTENSION: &str = "blob";
 
--- a/src/storage/core.rs
+++ b/src/storage/core.rs
@@ -653,9 +653,6 @@ where
                 },
                 _ => false,
             };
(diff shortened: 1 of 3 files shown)


src/storage/core.rs
```diff
@@ -649,9 +650,12 @@ where
                 ErrorKind::Bincode(_) => true,
                 ErrorKind::Validation { kind, cause: _ } => {
                     !matches!(kind, ValidationErrorKind::BlobVersion)
-                }
+                },
                 _ => false,
             };
+        } else if let Some(bincode::ErrorKind::Io(error)) =
+            error.downcast_ref::<bincode::Error>().and_then(|b| Some(b.as_ref())) {
+            return error.kind() == std::io::ErrorKind::UnexpectedEof;
```
ikopylov, line 658: You are trying to detect the error that is not correspond to the current level of abstraction. From the arhitectural point of view it is much better to detect this error at the place, where it happened and convert it to another error from our own `ErrorKind` (it seems that the error can be converted to `ErrorKind::Bincode`)

ikopylov:
This request should be backported to an earlier version of pearl (concrete version will be determined later)

ikopylov:
@Boneyan  Have you rechecked that after the last changes corrupted BLOB is detected correctly?

Boneyan:
Yes

ikopylov (approved):
LGTM

ikopylov:
@Boneyan please, create backport request to `release_v0.9.3`

PR #161, opened 2022-04-05
159 remove active cycle

### LINKED ISSUE #159: Remove active actions in observer cycle
Fixes https://github.com/qoollo/bob/issues/454

idruzhitskiy:
closes #159

### DIFF
--- a/src/storage/config.rs
+++ b/src/storage/config.rs
@@ -7,7 +7,7 @@ pub(crate) struct Config {
     max_blob_size: Option<u64>,
     max_data_in_blob: Option<u64>,
     blob_file_name_prefix: Option<String>,
-    update_interval_ms: u64,
+    debounce_interval_ms: u64,
     allow_duplicates: bool,
     ignore_corrupted: bool,
     index: IndexConfig,
@@ -39,8 +39,8 @@ impl Config {
     }
 
     #[inline]
-    pub const fn update_interval_ms(&self) -> u64 {
-        self.update_interval_ms
+    pub const fn debounce_interval_ms(&self) -> u64 {
+        self.debounce_interval_ms
     }
 
     #[inline]
@@ -134,7 +134,7 @@ impl Default for Config {
             max_blob_size: None,
             max_data_in_blob: None,
             blob_file_name_prefix: None,
-            update_interval_ms: 100,
+            debounce_interval_ms: 100,
             allow_duplicates: false,
             ignore_corrupted: false,
             index: Default::default(),
--- a/src/storage/config.rs
+++ b/src/storage/config.rs
@@ -134,7 +134,7 @@ impl Default for Config {
             max_blob_size: None,
             max_data_in_blob: None,
             blob_file_name_prefix: None,
-            debounce_interval_ms: 100,
+            debounce_interval_ms: 200,
             allow_duplicates: false,
             ignore_corrupted: false,
             index: Default::default(),
--- a/src/storage/core.rs
+++ b/src/storage/core.rs
@@ -291,7 +291,7 @@ where
             .active_blob
             .as_mut()
             .ok_or_else(Error::active_blob_not_set)?;
-        blob.write(record).await.or_else(|err| {
+        let result = blob.write(record).await.or_else(|err| {
             let e = err.downcast::<Error>()?;
             if let ErrorKind::FileUnavailable(kind) = e.kind() {
                 let work_dir = self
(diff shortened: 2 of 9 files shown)


src/storage/core.rs
```diff
@@ -303,8 +303,35 @@ where
...
+            .config
+            .max_data_in_blob()
+            .ok_or_else(|| Error::from(ErrorKind::Uninitialized))?;
+        if active_blob.file_size() > config_max_size
+            || active_blob.records_count() as u64 > config_max_count
+        {
+            let now = std::time::SystemTime::now();
+            let dur = now.duration_since(active_blob.created_at())?;
+            if dur.as_millis() > self.inner.config.debounce_interval_ms() as u128 {
+                self.observer.force_update_active_blob_always().await;
```
ikopylov, line 328: Force update will lead to a bug: successive put requests will generate many `OperationType::ForceUpdateActiveBlob` messages before the new BLOB will actually be created and thus many BLOBs will be created in sequence. You should emit something like `TryUpdateActiveBlobByThresholds` and during the processing of that message all conditions should be rechecked

src/storage/core.rs
```diff
@@ -303,8 +303,35 @@ where
...
+                self.observer.try_dump_old_blob_indexes().await;
```
ikopylov, line 329: Index dumping should be called within `TryUpdateActiveBlobByThresholds` processing code

src/storage/core.rs
```diff
@@ -303,8 +303,35 @@ where
...
+            .inner
+            .config
+            .max_data_in_blob()
+            .ok_or_else(|| Error::from(ErrorKind::Uninitialized))?;
+        if active_blob.file_size() > config_max_size
+            || active_blob.records_count() as u64 > config_max_count
+        {
+            let now = std::time::SystemTime::now();
+            let dur = now.duration_since(active_blob.created_at())?;
+            if dur.as_millis() > self.inner.config.debounce_interval_ms() as u128 {
```
ikopylov, line 327: SystemTime can easily be changed by user, so `dur.as_millis()` can easily be negative. I suggest to use `abs` when comparing

src/storage/core.rs
```diff
@@ -303,8 +303,35 @@ where
...
+            .config
+            .max_blob_size()
+            .ok_or_else(|| Error::from(ErrorKind::Uninitialized))?;
+        let config_max_count = self
+            .inner
+            .config
+            .max_data_in_blob()
+            .ok_or_else(|| Error::from(ErrorKind::Uninitialized))?;
+        if active_blob.file_size() > config_max_size
+            || active_blob.records_count() as u64 > config_max_count
```
ikopylov, line 323: `>=`

src/storage/config.rs
```diff
@@ -134,7 +134,7 @@ impl Default for Config {
             max_blob_size: None,
             max_data_in_blob: None,
             blob_file_name_prefix: None,
-            update_interval_ms: 100,
+            debounce_interval_ms: 100,
```
ikopylov, line 137: Let's increase this interval to `200`

src/storage/observer_worker.rs
```diff
@@ -87,6 +78,12 @@ where
                     .try_dump_old_blob_indexes(self.dump_sem.clone())
                     .await;
             }
+            OperationType::TryUpdateActiveBlob => {
+                self.try_update_active_blob().await?;
+                self.inner
```
ikopylov, line 83: I think, `try_dump_old_blob_indexes` only needs to be executed if the active blob has been updated.

Additional question: what if the active blob will never be updated (new timestamp with separate pearl instance will be created earlier)? Is there any background procedure that dumps old blob indices?

idruzhitskiy: Changed dump to work only in case of active blob checks.
Regarding blob never updating - no, there are no background processes for dumping indexes.

src/storage/observer_worker.rs
```diff
@@ -99,56 +97,62 @@ where
...
-        } else {
-            // if active blob doesn't exists, it doesn't need to be updated
-            return Ok(None);
+        let mut write = self.inner.safe.write().await;
+        let active_blob = write.active_blob.as_ref();
+        if let Some(active_blob) = active_blob {
+            if active_blob.file_size() >= config_max_size
+                || active_blob.records_count() as u64 >= config_max_count
+            {
+                let next_name = self.inner.next_blob_name()?;
```
ikopylov, line 130: You have `update_active_blob` function below. Why not call it from here?

idruzhitskiy: We have lock already taken here, and update function takes lock later

idruzhitskiy: It also means more actions during file lock (obtaining name and creating actual blob file) than before. It does not create any noticeable overhead during tests.

idruzhitskiy: Oh, we don't need complex lock logic here since it can be accessed from single thread only. You're right, update blob can be reused

src/storage/observer_worker.rs
```diff
@@ -99,56 +97,41 @@ where
...
-        .ok_or_else(|| Error::from(ErrorKind::Uninitialized))?;
-    let config_max_count = inner
-        .config
-        .max_data_in_blob()
-        .ok_or_else(|| Error::from(ErrorKind::Uninitialized))?;
-    if active_size as u64 > config_max_size || active_count >= config_max_count {
-        Ok(Some(inner))
-    } else {
-        Ok(None)
+        update_active_blob(&self.inner).await.map(|_| true)
```
ikopylov, line 123: This update should be done inside `write` lock with limit checks (as it was before the last commit). My previous comment was only about the logic after all the checks

idruzhitskiy: Why? It can't be updated elsewhere, so there is no need to keep write lock

ikopylov: Here is the public method, that updates acitve blob directly (without observer): https://github.com/qoollo/pearl/blob/master/src/storage/core.rs#L178

idruzhitskiy: Ah, since blob can be missing now we have synchronous operations... You're right, write lock should be held.

ikopylov (approved):
LGTM

PR #167, opened 2022-04-21
166 fix delete

idruzhitskiy:
Closes #166

PR #168, opened 2022-04-22
165 defer index dump on delete

### LINKED ISSUE #165: Defer index dump on record delete
Closes https://github.com/qoollo/bob/issues/363

idruzhitskiy:
Closes #165

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Pearl changelog
 
 ## [Unreleased]
 #### Added
+Add support for hierarchical range filters (#154)
 
 
 #### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,13 @@ Pearl changelog
 
 ## [Unreleased]
 #### Added
+- Add method to receive occupied disk space (#170)
 
 #### Changed
 - Defer index drop on delete (#165)
 
 #### Fixed
+- Restore documentation publishing (#173)
 
 
 #### Updated
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Pearl changelog
 #### Changed
 
 #### Fixed
+- Fix delete with no active blob set
 
 #### Updated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,14 +4,23 @@ Pearl changelog
 
 ## [Unreleased]
 #### Added
-Add support for hierarchical range filters (#154)
 
+#### Changed
+
+#### Fixed
+
+#### Updated
+
+
+## [0.11.0] - 2022-04-18
+#### Added
+Add support for hierarchical range filters (#154)
 
 #### Changed
 - Remove active checking for blob update (#159)
 
 #### Fixed
-- corrupted blob now should be saved in case of 'unexpected eof' error
+- corrupted blob now should be saved in case of 'unexpected eof' error (#160)
 
 #### Updated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Pearl changelog
 #### Added
 
 #### Changed
+- Defer index drop on delete (#165)
 
 #### Fixed
 - Fix delete with no active blob set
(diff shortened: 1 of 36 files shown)


CHANGELOG.md
```diff
@@ -6,8 +6,10 @@ Pearl changelog
 #### Added
 
 #### Changed
+- Defer index drop on delete (#165)
 
 #### Fixed
+- Fix delete with no active blob set
```
ikopylov, line 12: This was fixed here: https://github.com/qoollo/pearl/pull/167

src/storage/observer_worker.rs
```diff
@@ -85,6 +107,37 @@ where
...
+        let mut deferred = self.deferred.write().await;
+        if let Some(ref mut deferred) = *deferred {
+            deferred.update_last_time();
+        } else {
+            let sender = self.loopback_sender.clone();
+            let data = DeferredEventData::new();
+            *deferred = Some(data);
+            let min = self.deferred_min_duration.clone();
+            let max = self.deferred_max_duration.clone();
+            let deferred = self.deferred.clone();
```
ikopylov, line 125: `deferred` redefined here, so it is not easy to find out what `deferred` is moved into new thread. I suggest to give a different name here

src/blob/core.rs
```diff
@@ -285,9 +285,6 @@ where
             let record = Record::deleted(key)?;
             self.write(record).await?;
             let res = self.index.mark_all_as_deleted(key)?;
-            if on_disk {
```
ikopylov, line 288: We should be able to dump immediately if we are low on memory. I suggest to pass additional boolean flag `dump_immediately`, that controls this behavior and on the Bob side set it according to `index_memory_limit` to avoid exceeding it

idruzhitskiy: Isn't it better to leave it to bob? We can add method like `trim` or `free_possible_resources` to pearl public API and make bob use it if limits are reached. Adding flag for only one operation seems like ad hoc solution

ikopylov: On a call, we decided to add a flag to the function

ikopylov (approved):
LGTM

src/storage/core.rs
```diff
@@ -267,6 +267,26 @@ where
         self.write_with_optional_meta(key, value, Some(meta)).await
     }
 
+    /// Free all resources that may be freed without work interruption
+    /// NOTICE! This function frees part of the resources in separate thread,
+    /// so actual resources may be freed later
+    pub async fn free_freeable_resources(&self) {
+        self.observer.try_dump_old_blob_indexes().await;
+        self.close_active_blob_in_background().await;
```
ikopylov, line 275: This is probably incorrect. We should free only resources, that is optional and was loaded while performing record deletion. Closing an active blob is a visible change and will lead to update of active blob every time, the deletion happened

src/storage/core.rs
```diff
@@ -267,6 +267,26 @@ where
...
+
+    /// Get size in bytes of inactive indexes
+    pub async fn inactive_index_memory(&self) -> usize {
+        let safe = self.inner.safe.read().await;
+        let blobs = safe.blobs.read().await;
+        blobs.iter().fold(0, |s, n| s + n.index_memory())
+    }
+
+    /// Get size in bytes of all freeable resources
+    pub async fn freeable_resources_memory(&self) -> usize {
```
ikopylov, line 286: `index_memory` is a better name. The function with that name was previously used by `index_memory` metric, so it is good idea to keep the name, but take index memory from inactive blobs into account

src/storage/core.rs
```diff
@@ -267,6 +267,26 @@ where
         self.write_with_optional_meta(key, value, Some(meta)).await
     }
 
+    /// Free all resources that may be freed without work interruption
+    /// NOTICE! This function frees part of the resources in separate thread,
+    /// so actual resources may be freed later
+    pub async fn free_freeable_resources(&self) {
```
ikopylov, line 273: Maybe `free_excess_resources`?

ikopylov:
`tests` are failed. Please, fix them

src/storage/core.rs
ikopylov: In addition, it can return the amount of memory that has been freed

ikopylov (approved):
LGTM

PR #179, opened 2022-08-15
Index validation fix

Boneyan:
Correct blob size comparison in index validation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Pearl changelog
 
 
 #### Fixed
-- Correct blob size comparison in index validation
+- Correct blob size comparison in index validation (#179)
 
 #### Updated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Pearl changelog
 
 
 #### Fixed
-
+- Correct blob size comparison in index validation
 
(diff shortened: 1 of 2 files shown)


src/tools/validation.rs
```diff
@@ -20,7 +20,11 @@ where
 
     static_assertions::const_assert_eq!(HEADER_VERSION, 5);
     let header = read_index_header(path)?;
-    let blob_size = path.metadata()?.len();
+    let blob_path = path.with_extension("blob");
+    if !blob_path.exists() {
+        return Err(Error::index_header_validation_error("blob file doesn't exist").into());
```
ikopylov, line 25: This should not be an error. This should be a warning and the validation should be started anyway and obviously should not fail on invalid blob size (for simplicity, you can pass size from the index header if blob is not exists)

CHANGELOG.md
```diff
@@ -10,7 +10,7 @@ Pearl changelog
 
 
 #### Fixed
-
+- Correct blob size comparison in index validation
```
ikopylov, line 13: Add number of issue or pull request (in this case) inside braces
```suggestion
- Correct blob size comparison in index validation (#179)
```

ikopylov (approved):
LGTM

PR #180, opened 2022-08-21
Strict blob validation

Boneyan:
Returning error when validating blob with corrupted records

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Pearl changelog
 
 
 #### Fixed
-
+- Returning error when validating blob with corrupted records
 
 #### Updated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Pearl changelog
 
 
 #### Fixed
-- Returning error when validating blob with corrupted records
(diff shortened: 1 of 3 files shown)


src/tools/validation.rs
```diff
@@ -50,7 +50,7 @@ pub fn validate_blob(path: &Path) -> AnyResult<()> {
     let mut reader = BlobReader::from_path(&path)?;
     reader.read_header()?;
     while !reader.is_eof() {
-        reader.read_record(false)?;
+        reader.read_record(true)?;
```
ikopylov, line 53: You should fix `read_record` instead. It looks like `if skip_wrong` condition at the beggining of `read_record` should be inversed. Additionally, validate for correctness all places in Pearl and Bob, where `read_record` is called

CHANGELOG.md
```diff
@@ -10,7 +10,7 @@ Pearl changelog
 
 
 #### Fixed
-
+- Returning error when validating blob with corrupted records
```
ikopylov, line 13: Add a number of issue or pull request inside brackets
```suggestion
- Returning error when validating blob with corrupted records (#180)
```

ikopylov (approved):
LGTM

PR #182, opened 2022-08-26
181 fix records order bptree

### LINKED ISSUE #181: Order of records in bptree after deserialization is reversed
https://github.com/qoollo/pearl/blob/master/src/blob/index/core.rs#L311

Right now after deserialization indexes have records reversed which leads to returning oldest record instead of newest

idruzhitskiy:
Closes #181

### DIFF
--- a/src/blob/index/bptree/core.rs
+++ b/src/blob/index/bptree/core.rs
@@ -131,6 +131,12 @@ where
                 }
                 Ok(headers)
             })
+            .map(|mut headers| {
+                for val in headers.values_mut() {
+                    val.reverse();
+                }
+                headers
+            })
             .map(|headers| (headers, self.header.records_count))
     }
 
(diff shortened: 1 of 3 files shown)


src/blob/index/bptree/core.rs
```diff
@@ -131,6 +131,12 @@ where
                 }
                 Ok(headers)
             })
+            .map(|mut headers| {
+                for val in headers.values_mut() {
+                    val.reverse();
```
ikopylov, line 136: For the most common case when `val.len() == 1` we can avoid calling `reverse`. Please, add condition before the call

idruzhitskiy: I'm surprised this check is not part of the reverse itself. Ok, I'll add this

ikopylov: According to the source code (https://doc.rust-lang.org/src/core/slice/mod.rs.html#678) it is not a part

ikopylov (approved):
LGTM

PR #196, opened 2022-09-04
Add file name to error messages in Blob struct

### LINKED ISSUE #183: Add file name to error messages
This should simplify problems detection.
Here: https://github.com/qoollo/pearl/blob/master/src/blob/core.rs#L326
And in other placess

ikopylov:
Closes #183

### DIFF
--- a/src/blob/core.rs
+++ b/src/blob/core.rs
@@ -92,12 +92,12 @@ where
         } else {
             self.fsyncdata()
                 .await
-                .with_context(|| format!("Blob file dump failed! Blob file: {:?}", self.name.to_path()))?;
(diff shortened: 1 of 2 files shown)


src/blob/core.rs
```diff
@@ -92,12 +92,12 @@ where
         } else {
             self.fsyncdata()
                 .await
-                .with_context(|| "Blob file dump failed!")?;
+                .with_context(|| format!("Blob file dump failed! Blob file: {:?}", self.name.to_path()))?;
```
idruzhitskiy, line 95: Other errors start with lowercase, probably this one should be lowercase too. And also exclamation mark can be ommited in favor of comma IMO

idruzhitskiy (commented):
LGTM

PR #198, opened 2022-09-06
187 update locks

### LINKED ISSUE #187: Review and improve locks in `Storage`
`RWLock` on `safe` does protect `active_blob` + `blobs` updates inside, so the `write` lock should be acquired only when these 2 fields updating. But if we look at `write_with_optional_meta` function, it is incorrectly acquire `write` lock (https://github.com/qoollo/pearl/blob/master/src/storage/core.rs#L313). It should acquire `read` lock, because it does not update neither `active_blob` nor `blobs` fields.
Same thing here: https://github.com/qoollo/pearl/blob/master/src/storage/core.rs#L890 and here: https://github.com/qoollo/pearl/blob/master/src/storage/core.rs#L912. Probably, there are more places where `write` lock taken incorrectly.

Also, it is strange that `blobs` protected with additional `RWLock`. In every places this lock acquired right after acquire of `RWLock` on `safe`, thus can be removed.

This incorrect locking mechanism results in significant performance degradation

idruzhitskiy:
Closes #187

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Pearl changelog
 
 ## [Unreleased]
 #### Added
+- Add functions to get info about blobs and indexes (#151)
+- Add key_size field to index header (#151)
 
 
 #### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Pearl changelog
 - Returning error when validating blob with corrupted records (#180)
 
 #### Updated
+- Update locks (#187)
 
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,11 @@ Pearl changelog
 
 
 #### Changed
+- File name added to error messages in Blob struct (#183)
 - Update rust edition (#175)
 
 #### Fixed
+- Fix docs (#177)
 - Fix order of records after bptree deserialization (#181)
 - Correct blob size comparison in index validation (#179)
 - Returning error when validating blob with corrupted records (#180)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,20 @@ Pearl changelog
 
 
 #### Changed
+
+
+#### Fixed
+
+
+#### Updated
+- Updated tokio, nix, clap, bitvec versions (#202)
+
+
+## [0.14.0] - 2022-11-14
+#### Added
+
+
+#### Changed
 - File name added to error messages in Blob struct (#183)
 - Update rust edition (#175)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Pearl changelog
 
 #### Updated
 - Updated tokio, nix, clap, bitvec versions (#202)
+- Update locks (#187)
 
 
 ## [0.14.0] - 2022-11-14
@@ -31,7 +32,6 @@ Pearl changelog
 - Returning error when validating blob with corrupted records (#180)
 
 #### Updated
-- Update locks (#187)
 
 
 
--- a/src/blob/core.rs
+++ b/src/blob/core.rs
@@ -21,7 +21,7 @@ where
     for<'a> K: Key<'a>,
 {
     header: Header,
-    index: Arc<RwLock<Index<K>>>,
+    index: Index<K>,
     name: FileName,
     file: File,
     current_offset: Arc<Mutex<u64>>,
@@ -50,7 +50,7 @@ where
         let header = Header::new();
         let mut blob = Self {
             header,
-            index: Arc::new(RwLock::new(index)),
+            index,
             name,
             file,
             current_offset,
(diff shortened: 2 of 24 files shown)


src/blob/core.rs
```diff
@@ -282,15 +289,17 @@ where
         }))
     }
 
-    pub(crate) async fn mark_all_as_deleted(&mut self, key: &K) -> Result<Option<u64>> {
-        if self.index.get_any(key).await?.is_some() {
-            let on_disk = self.index.on_disk();
+    pub(crate) async fn mark_all_as_deleted(blob: &ASRwLock<Self>, key: &K) -> Result<Option<u64>> {
```
ikopylov, line 292: The lock in the new implementation is acquired multiple times, which can lead to performance degradation. The original implementation was better. The only possible improvement here can be achieved by acquiring `upgradable_read` lock at first, then check key presence (`write.index.get_any(key).await?.is_some()`) and then upgrade to write lock if it is presented (any further lock manipulations should be avoided)

src/lib.rs
```diff
@@ -46,6 +46,8 @@ extern crate anyhow;
 
 extern crate ring;
 
+extern crate async_std;
```
ikopylov, line 49: Why is it exported? Normally, this crate should be purely internal, as it is an implementation detail

src/storage/core.rs
```diff
@@ -897,30 +901,33 @@ where
...
-        let total = entries_closed_blobs
-            .filter_map(|result| match result {
-                Ok(count) => count,
-                Err(error) => {
-                    warn!("failed to delete records: {}", error);
-                    None
+        let count = AtomicU64::new(0);
+        blobs
+            .for_each(|b| async {
+                let rwlock = ASRwLock::new(b);
```
ikopylov, line 907: Oh, it is not good to wrap blob with RwLock only in order to use the new signature of `Blob::mark_all_as_deleted`. I think, it is better just to revert to the original implementation, where `mark_all_as_deleted` just received `Blob` as `self`

src/blob/core.rs
ikopylov: I just saw a removal implementation in `src/storage/core.rs`. I think, it is better just to revert to the original form of this function

src/storage/core.rs
```diff
@@ -897,30 +901,33 @@ where
...
+                let result = Blob::mark_all_as_deleted(&rwlock, key).await;
+                match result {
+                    Ok(cnt) => {
+                        count.fetch_add(cnt.unwrap_or(0), Ordering::Release);
+                    }
+                    Err(err) => warn!("failed to delete records: {}", err),
```
ikopylov, line 913: Can you check why this error is not propagated above?

ikopylov (changes requested):
Changes are required

src/storage/core.rs
ikopylov: It would be good to return error to the Bob client if the error happened somewhere inside. But we should not stop processing other blobs if error happened in one

idruzhitskiy: Yep, error is returned if any blobs contained error, but processing is not stopped

src/filter/hierarchical.rs
```diff
@@ -423,6 +436,26 @@ impl<Key, Filter, Child> HierarchicalFilters<Key, Filter, Child> {
...
     }
 
+    pub(crate) async fn for_each<F, R>(&mut self, f: F)
+    where
+        F: Fn(Child) -> R + Copy,
+        R: futures::Future<Output = Child>,
+    {
+        self.children = self
+            .children
+            .drain(..)
```
ikopylov, line 446: According to documentation `drain` will clear `children` collection. Looks like an error

src/filter/hierarchical.rs
```diff
@@ -423,6 +436,26 @@ impl<Key, Filter, Child> HierarchicalFilters<Key, Filter, Child> {
...
+            .map(|o| async move {
+                if let Some(c) = o {
+                    Some(c.map_async(f).await)
+                } else {
+                    None
+                }
+            })
+            .collect::<FuturesUnordered<_>>()
+            .collect::<Vec<_>>()
```
ikopylov, line 455: The only place, where `for_each` is used just ignores the result of this function. So it is better to not collect results into a Vector

src/storage/core.rs
```diff
@@ -897,29 +901,39 @@ where
...
-            .iter_mut()
-            .map(|b| b.mark_all_as_deleted(key))
-            .collect::<FuturesUnordered<_>>();
-        let total = entries_closed_blobs
-            .filter_map(|result| match result {
-                Ok(count) => count,
-                Err(error) => {
-                    warn!("failed to delete records: {}", error);
-                    None
+        let count = AtomicU64::new(0);
```
ikopylov, line 904: IMHO, original implementation of this function without Atomics was clearer and simpler

src/blob/core.rs
```diff
@@ -229,11 +235,16 @@ where
         // @TODO implement
     }
 
-    pub(crate) async fn write(&mut self, mut record: Record) -> Result<()> {
+    pub(crate) async fn write(blob: &ASRwLock<Self>, record: Record) -> Result<()> {
+        debug!("blob write");
+        let blob = blob.upgradable_read().await;
```
ikopylov, line 240: Please, add comment that only one thread can acquire an `upgradable_read` lock. Because this is not obvious thing and the further code depends on this behavior (only one thread can write data at `self.current_offset`)

src/filter/hierarchical.rs
idruzhitskiy: This is done to allow creation of futures from each of the values and processing it in parallel

src/filter/hierarchical.rs
idruzhitskiy: This is needed to recreate source vector after parallel processing

src/filter/hierarchical.rs
idruzhitskiy: We recreate leaves with the new data

src/storage/core.rs
idruzhitskiy: You right, since we don't need ownership here I'll return old function

src/filter/hierarchical.rs
idruzhitskiy: Function removed

src/filter/hierarchical.rs
idruzhitskiy: This was to allow for ownership transfer into closure. Now function is removed

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ Pearl changelog
 - Returning error when validating blob with corrupted records (#180)
 
 #### Updated
+- Update locks (#187)
```
ikopylov, line 34: Please, move this line to `[Unreleased]` section

ikopylov (approved):
LGTM

PR #203, opened 2022-09-13
193 fix excessive conversion and add impls where needed

### LINKED ISSUE #193: The key is converted to a vector and back to a key two times in `IndexStruct::push`
First strange conversion: https://github.com/qoollo/pearl/blob/master/src/blob/index/core.rs#L268
Second: https://github.com/qoollo/pearl/blob/master/src/blob/index/core.rs#L275

Looks like `h.key()` can be used directly

lostl1ght:
Closes #193

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Pearl changelog
 - Fix order of records after bptree deserialization (#181)
 - Correct blob size comparison in index validation (#179)
 - Returning error when validating blob with corrupted records (#180)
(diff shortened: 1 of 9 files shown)


ikopylov:
@lostl1ght Please, fix conflicts. Key trait was moved to a separate file. Also, there was `ArrayKey<>` struct added

lostl1ght:
@ikopylov Done.

ikopylov (approved):
LGTM

PR #204, opened 2022-09-14
194 remove iterators from bloom filter

### LINKED ISSUE #194: Avoid using iterators in Bloom filter on hot paths
Bloom filter is heavily used in every read and write operation, so its methods should be optimized as much as possible.
I suggest to replace iterators by simple `for` loops in `add` (https://github.com/qoollo/pearl/blob/master/src/filter/bloom.rs#L269) and `contains_in_memory` (https://github.com/qoollo/pearl/blob/master/src/filter/bloom.rs#L281) functions

lostl1ght:
Closes #194

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Pearl changelog
 #### Changed
 - File name added to error messages in Blob struct (#183)
 - Update rust edition (#175)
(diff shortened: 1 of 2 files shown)


ikopylov (approved):
LGTM

PR #209, opened 2022-09-19
207 preserve delete

### LINKED ISSUE #207: Preserve deletion record in index
Required for https://github.com/qoollo/bob/issues/533

idruzhitskiy:
Closes  #207

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Pearl changelog
 
 
 #### Changed
+- Save delete records in index (#207)
 
 
 #### Fixed
@@ -41,7 +42,6 @@ Pearl changelog
 #### Changed
 - File name added to error messages in Blob struct (#183)
 - Update rust edition (#175)
-- Save delete records in index (#207)
 
 #### Fixed
 - Fix docs (#177)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Pearl changelog
 #### Changed
 - File name added to error messages in Blob struct (#183)
 - Update rust edition (#175)
+- Save delete records in index (#207)
 
 #### Fixed
 - Fix docs (#177)
--- a/src/blob/core.rs
+++ b/src/blob/core.rs
@@ -216,7 +216,6 @@ where
                 self.name.to_path()
             )
         })? {
-            let headers = filter_deleted_headers(headers);
             for header in headers {
                 self.index.push(header).context("index push failed")?;
             }
@@ -531,27 +530,6 @@ impl RawRecords {
         self.current_offset += header.data_size();
         Ok(header)
     }
-}
-
-pub(crate) fn filter_deleted_headers(headers: Vec<RecordHeader>) -> Vec<RecordHeader> {
-    let deleted = headers.iter().fold(BTreeMap::new(), |mut map, h| {
-        if h.is_deleted() {
-            let entry = map.entry(h.key().to_vec()).or_insert(h.created());
-            *entry = h.created().max(*entry);
-        }
-        map
-    });
-    if deleted.is_empty() {
-        return headers;
-    }
-
-    let mut new_headers = vec![];
-    for header in headers {
-        if deleted.get(header.key()).cloned().unwrap_or_default() < header.created() {
-            new_headers.push(header);
-        }
-    }
-    new_headers
 }
 
 #[async_trait::async_trait]
--- a/src/blob/core.rs
+++ b/src/blob/core.rs
@@ -260,8 +260,9 @@ where
             })?;
         self.index.push(key, record.header().clone())?;
         self.current_offset += buf.len() as u64;
-        Ok(record.header().clone())
+        Ok(record.into_header())
     }
+
     async fn write_locked(
         blob: ASRwLockUpgradableReadGuard<'_, Blob<K>>,
         key: &K,
--- a/src/blob/core.rs
+++ b/src/blob/core.rs
@@ -241,7 +241,7 @@ where
         Self::write_locked(blob, record).await
     }
 
-    async fn write_mut(&mut self, mut record: Record) -> Result<()> {
+    async fn write_mut(&mut self, mut record: Record) -> Result<RecordHeader> {
         debug!("blob write");
         debug!("blob write record offset: {}", self.current_offset);
         record.set_offset(self.current_offset)?;
@@ -259,7 +259,7 @@ where
             })?;
         self.index.push(record.header().clone())?;
         self.current_offset += buf.len() as u64;
-        Ok(())
+        Ok(record.header().clone())
     }
     async fn write_locked(
         blob: ASRwLockUpgradableReadGuard<'_, Blob<K>>,
@@ -322,14 +322,21 @@ where
         }))
     }
 
-    pub(crate) async fn mark_all_as_deleted(&mut self, key: &K) -> Result<u64> {
+    pub(crate) async fn mark_all_as_deleted(&mut self, key: &K) -> Result<Option<u64>> {
+        if self.index.get_any(key).await?.is_some() {
+            Ok(Some(self.push_deletion_record(key).await?))
+        } else {
+            Ok(None)
+        }
+    }
+
+    async fn push_deletion_record(&mut self, key: &K) -> Result<u64> {
         let on_disk = self.index.on_disk();
         if on_disk {
             self.load_index().await?;
         }
         let record = Record::deleted(key)?;
-        let header = record.header.clone();
-        self.write_mut(record).await?;
+        let header = self.write_mut(record).await?;
         let res = self.index.mark_all_as_deleted(key, header)?;
         Ok(res)
     }
--- a/src/blob/core.rs
+++ b/src/blob/core.rs
@@ -340,7 +340,7 @@ where
         }
         let record = Record::deleted(key)?;
         let header = self.write_mut(key, record).await?;
-        self.index.mark_all_as_deleted(key, header)
+        self.index.push_deletion(key, header)
     }
 
     fn headers_to_entries(headers: Vec<RecordHeader>, file: &File) -> Vec<Entry> {
--- a/src/blob/core.rs
+++ b/src/blob/core.rs
@@ -325,11 +325,12 @@ where
         }))
     }
 
-    pub(crate) async fn mark_all_as_deleted(&mut self, key: &K) -> Result<Option<()>> {
+    pub(crate) async fn mark_all_as_deleted(&mut self, key: &K) -> Result<bool> {
         if self.index.get_any(key).await?.is_some() {
-            Ok(Some(self.push_deletion_record(key).await?))
+            self.push_deletion_record(key).await?;
+            Ok(true)
         } else {
-            Ok(None)
+            Ok(false)
         }
     }
 
--- a/src/blob/core.rs
+++ b/src/blob/core.rs
@@ -223,7 +223,8 @@ where
             )
         })? {
             for header in headers {
-                self.index.push(header).context("index push failed")?;
+                let key = header.key().into();
+                self.index.push(&key, header).context("index push failed")?;
             }
         }
         debug!("index successfully generated: {}", self.index.name());
@@ -234,14 +235,14 @@ where
         // @TODO implement
     }
 
-    pub(crate) async fn write(blob: &ASRwLock<Self>, record: Record) -> Result<()> {
+    pub(crate) async fn write(blob: &ASRwLock<Self>, key: &K, record: Record) -> Result<()> {
         debug!("blob write");
         // Only one upgradable_read lock is allowed at a time
         let blob = blob.upgradable_read().await;
-        Self::write_locked(blob, record).await
+        Self::write_locked(blob, key, record).await
     }
 
-    async fn write_mut(&mut self, mut record: Record) -> Result<RecordHeader> {
+    async fn write_mut(&mut self, key: &K, mut record: Record) -> Result<RecordHeader> {
         debug!("blob write");
         debug!("blob write record offset: {}", self.current_offset);
         record.set_offset(self.current_offset)?;
@@ -257,12 +258,13 @@ where
                     _ => e.into(),
                 }
             })?;
-        self.index.push(record.header().clone())?;
+        self.index.push(key, record.header().clone())?;
         self.current_offset += buf.len() as u64;
         Ok(record.header().clone())
     }
     async fn write_locked(
         blob: ASRwLockUpgradableReadGuard<'_, Blob<K>>,
+        key: &K,
         mut record: Record,
     ) -> Result<()> {
         debug!("blob write record offset: {}", blob.current_offset);
@@ -280,7 +282,7 @@ where
                 }
             })?;
         let mut blob = RwLockUpgradableReadGuard::upgrade(blob).await;
-        blob.index.push(record.header().clone())?;
+        blob.index.push(key, record.header().clone())?;
         blob.current_offset += buf.len() as u64;
         Ok(())
     }
@@ -322,7 +324,7 @@ where
         }))
     }
 
-    pub(crate) async fn mark_all_as_deleted(&mut self, key: &K) -> Result<Option<u64>> {
+    pub(crate) async fn mark_all_as_deleted(&mut self, key: &K) -> Result<Option<()>> {
         if self.index.get_any(key).await?.is_some() {
             Ok(Some(self.push_deletion_record(key).await?))
         } else {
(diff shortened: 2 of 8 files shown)


src/blob/core.rs
```diff
@@ -282,19 +295,16 @@ where
...
-            if on_disk {
-                self.load_index().await?;
-            }
-            let record = Record::deleted(key)?;
-            self.write(record).await?;
-            let res = self.index.mark_all_as_deleted(key)?;
-            Ok(res)
-        } else {
-            Ok(None)
+    pub(crate) async fn mark_all_as_deleted(&mut self, key: &K) -> Result<u64> {
```
ikopylov, line 298: I think, this function should be renamed. It is now adds a deletion record, so possible name can be `put_deletion_record`.
Also, it would be nice to keep the old `mark_all_as_deleted` function, which first checks for existence in the index and then calls the new one.

src/blob/index/core.rs
```diff
@@ -244,9 +244,21 @@ where
...
+    fn mark_all_as_deleted_in_memory(
+        headers: &mut InMemoryIndex<K>,
+        key: &K,
+        default_header: RecordHeader,
+    ) -> Result<u64> {
         debug!("headers: {}", headers.len());
-        headers.remove(key).map(|v| v.len() as u64)
+        if let Some(v) = headers.get_mut(key) {
+            for h in v.iter_mut() {
+                h.mark_as_deleted()?
```
ikopylov, line 255: This is not good. It is better to keep existing records as is and push deletion record to the end of vector

src/blob/index/core.rs
```diff
@@ -244,9 +244,21 @@ where
         }
     }
 
-    fn mark_all_as_deleted_in_memory(headers: &mut InMemoryIndex<K>, key: &K) -> Option<u64> {
+    fn mark_all_as_deleted_in_memory(
```
ikopylov, line 247: I think it is better to name function as `push_deletion_header_in_memory`

src/blob/core.rs
```diff
@@ -282,19 +295,16 @@ where
...
+        let on_disk = self.index.on_disk();
+        if on_disk {
+            self.load_index().await?;
         }
+        let record = Record::deleted(key)?;
+        let header = record.header.clone();
```
ikopylov, line 304: Looks like this `clone` should happen after the record is written

src/blob/index/mod.rs
```diff
@@ -33,5 +33,5 @@ pub(crate) trait IndexTrait<K>: Send + Sync {
     fn count(&self) -> usize;
     async fn dump(&mut self, blob_size: u64) -> Result<usize>;
     async fn load(&mut self, blob_size: u64) -> Result<()>;
-    fn mark_all_as_deleted(&mut self, key: &K) -> Result<Option<u64>>;
+    fn mark_all_as_deleted(&mut self, key: &K, default_header: RecordHeader) -> Result<u64>;
```
ikopylov, line 36: `default_header` is not a good name for parameter. It can be just `header` (`default` is misleading)

ikopylov: Question: here you pass both `key` and `header` as parameters, but in `push` function above only `header` is passed. Is it possible to add `key` argument to `push` function?

src/record.rs
```diff
@@ -292,8 +292,4 @@ impl Header {
     pub(crate) fn is_deleted(&self) -> bool {
         self.flags & DELETE_FLAG == DELETE_FLAG
     }
-
-    pub(crate) fn created(&self) -> u64 {
```
ikopylov, line 296: Why this was removed?

src/blob/index/core.rs
```diff
@@ -244,9 +244,21 @@ where
...
+        key: &K,
+        default_header: RecordHeader,
+    ) -> Result<u64> {
         debug!("headers: {}", headers.len());
-        headers.remove(key).map(|v| v.len() as u64)
+        if let Some(v) = headers.get_mut(key) {
+            for h in v.iter_mut() {
+                h.mark_as_deleted()?
+            }
+            Ok(v.len() as u64)
```
ikopylov, line 257: It is still `1`, not `v.len()`, because the latest record overrides the old ones

src/storage/core.rs
```diff
@@ -920,7 +920,7 @@ where
         let mut safe = self.inner.safe.write().await;
         let active_blob = safe.active_blob.as_deref_mut();
         let count = if let Some(active_blob) = active_blob {
-            let count = active_blob.mark_all_as_deleted(key).await?.unwrap_or(0);
+            let count = active_blob.mark_all_as_deleted(key).await?;
```
ikopylov, line 923: It would be good to rename functions here too (`push_deletion_record`)

ikopylov:
Please, resolve merge conflicts

src/blob/index/core.rs
```diff
@@ -244,9 +244,21 @@ where
...
+        } else {
+            headers.insert(key.clone(), vec![default_header]);
```
ikopylov, line 259: Check `push` implementation. It also updates `mem` struct. Looks like this also should be done here

src/blob/index/core.rs
```diff
@@ -305,19 +317,19 @@ where
...
             }
             State::OnDisk(findex) => {
                 debug!("index get any on disk");
-                let header = findex.get_any(key).await?;
-                Ok(header)
+                findex.get_any(key).await?
             }
-        }
+        };
+        Ok(result.and_then(|h| if h.is_deleted() { None } else { Some(h) }))
```
ikopylov, line 332: Function `get_all` above should also be updated in the same way. Please, check other functions: maybe they also should be updated

src/blob/index/mod.rs
idruzhitskiy: No, `push` is used during regeneration, and there are no keys at this moment.

src/record.rs
idruzhitskiy: Unused

idruzhitskiy: It was removed by mistake

CHANGELOG.md
```diff
@@ -41,6 +41,7 @@ Pearl changelog
 #### Changed
 - File name added to error messages in Blob struct (#183)
 - Update rust edition (#175)
+- Save delete records in index (#207)
```
ikopylov, line 44: Move this line to `[Unreleased]` section

src/blob/index/core.rs
```diff
@@ -244,9 +244,28 @@ where
...
     }
 
-    fn mark_all_as_deleted_in_memory(headers: &mut InMemoryIndex<K>, key: &K) -> Option<u64> {
+    fn push_deletion_header_in_memory(
+        mem: &mut MemoryAttrs,
+        headers: &mut InMemoryIndex<K>,
+        key: &K,
+        header: RecordHeader,
+    ) -> Result<u64> {
+        mem.records_count += 1;
```
ikopylov, line 253: I think it is better to do after the insertion happened (at the end of the function)

src/blob/index/core.rs
```diff
@@ -296,27 +315,41 @@ where
...
-        match &self.inner {
+        let res = match &self.inner {
             State::InMemory(headers) => Ok(headers.get(key).cloned()),
             State::OnDisk(findex) => findex.find_by_key(key).await,
-        }
+        };
+        res.map(|r| {
+            r.and_then(|mut hs| {
+                hs.sort_by_key(|h| h.created());
+                hs.iter().fold(None, |s, n| {
```
ikopylov, line 325: This is not optimal. It is better to find the last `None` item and remove it and everything before directly from the vector without recreating it.

src/blob/index/core.rs
```diff
@@ -344,10 +377,17 @@ where
         }
     }
 
-    fn mark_all_as_deleted(&mut self, key: &K) -> Result<Option<u64>> {
+    fn mark_all_as_deleted(&mut self, key: &K, header: RecordHeader) -> Result<u64> {
```
ikopylov, line 380: `push_deletion`?

src/blob/index/core.rs
```diff
@@ -244,9 +244,28 @@ where
         }
     }
 
-    fn mark_all_as_deleted_in_memory(headers: &mut InMemoryIndex<K>, key: &K) -> Option<u64> {
+    fn push_deletion_header_in_memory(
```
ikopylov, line 247: You should also update `filters` to make the deletion visible (this will be required for the next PR)

src/blob/index/core.rs
ikopylov: As I see, this function becomes identical to `push`, which is good. So it is better to call `push` here, but additionally check, that passed `header` has `deleted` flag and its record size is zero

src/storage/core.rs
```diff
@@ -905,13 +905,13 @@ where
             .collect::<FuturesUnordered<_>>();
         let total = entries_closed_blobs
             .filter_map(|result| match result {
-                Ok(count) => count,
+                Ok(count) => Some(count),
```
ikopylov, line 908: If you reverted other changes to this function, then this one should be reverted too.
Also the question: do you plan to start using new functions in a subsequent PR?

src/blob/index/core.rs
```diff
@@ -296,27 +315,41 @@ where
...
     async fn get_all(&self, key: &K) -> Result<Option<Vec<RecordHeader>>> {
-        match &self.inner {
+        let res = match &self.inner {
             State::InMemory(headers) => Ok(headers.get(key).cloned()),
             State::OnDisk(findex) => findex.find_by_key(key).await,
-        }
+        };
+        res.map(|r| {
+            r.and_then(|mut hs| {
+                hs.sort_by_key(|h| h.created());
```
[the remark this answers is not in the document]
ikopylov: Please, recheck that this is true for `OnDisk` index

src/storage/core.rs
idruzhitskiy: Which functions?

ikopylov: I mean `push_deletion_record` function

src/blob/core.rs
```diff
@@ -258,12 +258,13 @@ where
                     _ => e.into(),
                 }
             })?;
-        self.index.push(record.header().clone())?;
+        self.index.push(key, record.header().clone())?;
         self.current_offset += buf.len() as u64;
-        Ok(())
+        Ok(record.header().clone())
```
ikopylov, line 263: Does the `clone()` needed here? If I understand right: we own `record` so we can return header without cloning

src/blob/index/bptree/core.rs
```diff
@@ -293,8 +292,11 @@ where
         self.file.read_at(&mut buf[..buf_size], leaf_offset).await?;
         let rh_size = self.header.record_header_size;
         if let Some((header, offset)) = self.read_header_buf(&buf[..buf_size], key, rh_size)? {
-            let mut headers = vec![header];
-            self.go_left(&mut headers, &buf[..buf_size], offset).await?;
+            let mut headers = vec![];
```
ikopylov, line 295: Create with capacity. By default, the first `push` to the vector increases its capacity to 4 items. In our use case the most common number of items is `1`

src/blob/index/core.rs
```diff
@@ -344,15 +352,11 @@ where
         }
     }
 
-    fn mark_all_as_deleted(&mut self, key: &K) -> Result<Option<u64>> {
+    fn mark_all_as_deleted(&mut self, key: &K, header: RecordHeader) -> Result<()> {
```
ikopylov, line 355: Rename to `push_deletion`

src/blob/index/core.rs
```diff
@@ -296,27 +290,41 @@ where
     }
 
     async fn get_all(&self, key: &K) -> Result<Option<Vec<RecordHeader>>> {
-        match &self.inner {
+        let res = match &self.inner {
             State::InMemory(headers) => Ok(headers.get(key).cloned()),
```
ikopylov, line 294: In memory all records stored in order of addition: the latest record is at the end. So you need to reverse the order here

src/blob/index/core.rs
```diff
@@ -296,27 +290,41 @@ where
...
             State::OnDisk(findex) => findex.find_by_key(key).await,
-        }
+        };
+        res.map(|r| {
+            r.and_then(|mut hs| {
```
ikopylov, line 298: Please, do it with pattern matching `if` (without `map` and `and_then`):
```
if Ok(Some(hs)) = res {
...
}
```

src/blob/index/core.rs
```diff
@@ -296,27 +290,41 @@ where
...
+                let last_del = (hs.len() - hs.iter().rev().take_while(|h| !h.is_deleted()).count())
```
ikopylov, line 299: The records are ordered from newest to latest, so you should find the first deletion record.
Also there is `position` function on `iter` that do what you need.

src/blob/index/core.rs
```diff
@@ -296,27 +290,41 @@ where
...
+                    .saturating_sub(1);
+                if last_del > 0 {
+                    hs.drain(0..=last_del);
```
ikopylov, line 302: With the order from newest to latest you need to remove the tail of the vector. That can be done with `truncate` function

src/blob/index/core.rs
ikopylov: Before calling `reverse` please check, that the length of the vector is `> 1`

src/blob/index/bptree/core.rs
```diff
@@ -293,8 +292,11 @@ where
...
+            self.go_left(header.key(), &mut headers, &buf[..buf_size], offset)
+                .await?;
+            headers.reverse();
```
ikopylov, line 298: Call it only if `headers.len() > 1`

src/blob/index/mod.rs
```diff
@@ -28,10 +28,10 @@ mod prelude {
...
     async fn get_all(&self, key: &K) -> Result<Option<Vec<RecordHeader>>>;
     async fn get_any(&self, key: &K) -> Result<Option<RecordHeader>>;
-    fn push(&mut self, h: RecordHeader) -> Result<()>;
+    fn push(&mut self, key: &K, h: RecordHeader) -> Result<()>;
     async fn contains_key(&self, key: &K) -> Result<bool>;
     fn count(&self) -> usize;
     async fn dump(&mut self, blob_size: u64) -> Result<usize>;
     async fn load(&mut self, blob_size: u64) -> Result<()>;
-    fn mark_all_as_deleted(&mut self, key: &K) -> Result<Option<u64>>;
+    fn mark_all_as_deleted(&mut self, key: &K, default_header: RecordHeader) -> Result<()>;
```
ikopylov, line 36: `default_header` -> `header`

src/storage/core.rs
```diff
@@ -911,7 +911,7 @@ where
                     None
                 }
             })
-            .fold(0, |a, b| a + b)
+            .fold(0, |a, _| a + 1)
```
ikopylov, line 914: `count`?

src/storage/core.rs
```diff
@@ -923,9 +923,9 @@ where
         let active_blob = safe.active_blob.as_deref_mut();
         let count = if let Some(active_blob) = active_blob {
             let mut active_blob = active_blob.write().await;
-            let count = active_blob.mark_all_as_deleted(key).await?.unwrap_or(0);
-            debug!("{} deleted from active blob", count);
-            count
+            active_blob.mark_all_as_deleted(key).await?;
+            debug!("1 deleted from active blob");
+            1
```
ikopylov, line 928: `mark_all_as_deleted` can return `None` in which case the result should be `0`

ikopylov:
There is a warning about `created` function. Previously, I suggested to keep that function, but if its presence generates warning then it is better to remove it (like you did it before). Sorry

src/storage/core.rs
idruzhitskiy: It was buried under ambigous calls, fixed

src/storage/core.rs
```diff
@@ -903,27 +903,32 @@ where
...
-            .filter_map(|result| match result {
-                Ok(count) => count,
-                Err(error) => {
-                    warn!("failed to delete records: {}", error);
-                    None
-                }
-            })
-            .fold(0, |a, b| a + b)
+        let total =
+            futures::stream::StreamExt::count(entries_closed_blobs.filter_map(
```
ikopylov, line 907: Why is the call to `count` in this form? Why not `.count()` at the end of expression?

src/blob/core.rs
```diff
@@ -323,21 +325,24 @@ where
         }))
     }
 
-    pub(crate) async fn mark_all_as_deleted(&mut self, key: &K) -> Result<Option<u64>> {
+    pub(crate) async fn mark_all_as_deleted(&mut self, key: &K) -> Result<Option<()>> {
```
ikopylov, line 328: I just thought, that it is strange to return `Option<()>` to show whether the record was deleted or not. There is a better type for that: `bool`. I suggest to replace `Result<Option<()>>` to `Result<bool>`

src/storage/core.rs
idruzhitskiy: Ambiguity with other imported methods

ikopylov (approved):
LGTM

PR #210, opened 2022-09-19
208 return deletion type

### LINKED ISSUE #208: Return special type for deleted records
Required for https://github.com/qoollo/bob/issues/533

idruzhitskiy:
Closes #208


CHANGELOG.md
```diff
@@ -41,6 +41,7 @@ Pearl changelog
 #### Changed
 - File name added to error messages in Blob struct (#183)
 - Update rust edition (#175)
+- Return delete or not found as distinct results (#208)
```
ikopylov, line 44: Move this to `[Unreleased]` section

src/blob/core.rs
```diff
@@ -350,51 +358,61 @@ where
         key: &K,
         meta: Option<&Meta>,
         check_filters: bool,
-    ) -> Result<Option<Entry>> {
+    ) -> Result<ReadResult<Entry>> {
```
ikopylov, line 361: `read_all_entries` above should also return `ReadResult<>`:
1. At least one non-deleted record -> `ReadResult::Found()`
2. There are records for the key, but the last one is deletion record -> `ReadResult::Deleted()`
3. No records -> `ReadResult::NotFound`

src/blob/core.rs
```diff
@@ -350,51 +358,61 @@ where
...
+                    let entry = Entry::new(header, self.file.clone());
+                    debug!("blob, get any entry, bloom true no meta, entry found");
+                    entry
+                }))
         }
     }
 
-    async fn get_entry_with_meta(&self, key: &K, meta: &Meta) -> Result<Option<Entry>> {
+    async fn get_entry_with_meta(&self, key: &K, meta: &Meta) -> Result<ReadResult<Entry>> {
         let headers = self.index.get_all(key).await?;
```
ikopylov, line 387: Looks like `index.get_all()` should return `ReadResult<>` as well

src/blob/core.rs
```diff
@@ -350,51 +358,61 @@ where
...
             self.filter_entries(entries, meta).await
         } else {
-            Ok(None)
+            Ok(ReadResult::NotFound)
         }
     }
 
-    async fn filter_entries(&self, entries: Vec<Entry>, meta: &Meta) -> Result<Option<Entry>> {
+    async fn filter_entries(&self, entries: Vec<Entry>, meta: &Meta) -> Result<ReadResult<Entry>> {
         for mut entry in entries {
```
ikopylov, line 397: Looks like there was a bug. If I understand the general logic, the latest by creation time entry is the last one in the vector. If this is true, then the iteration should be done in reverse order

src/blob/core.rs
```diff
@@ -350,51 +358,61 @@ where
...
+                    ReadResult::Found(entry)
+                });
             }
         }
-        Ok(None)
+        Ok(ReadResult::NotFound)
     }
 
-    pub(crate) async fn contains(&self, key: &K, meta: Option<&Meta>) -> Result<bool> {
+    pub(crate) async fn contains(&self, key: &K, meta: Option<&Meta>) -> Result<ReadResult<u64>> {
```
ikopylov, line 409: This is not an obvious change. You must create a wrapper struct (or typedef) to emphasize that this `u64` is the creation timestamp

src/blob/core.rs
```diff
@@ -323,8 +327,12 @@ where
         }))
     }
 
-    pub(crate) async fn mark_all_as_deleted(&mut self, key: &K) -> Result<Option<u64>> {
-        if self.index.get_any(key).await?.is_some() {
+    pub(crate) async fn mark_all_as_deleted(
```
ikopylov, line 330: You've changed that function in https://github.com/qoollo/pearl/pull/209. Probably, you should merge https://github.com/qoollo/pearl/pull/209 into this branch

src/blob/index/core.rs
```diff
@@ -257,7 +257,7 @@ where
     for<'a> K: Key<'a>,
 {
     async fn contains_key(&self, key: &K) -> Result<bool> {
```
ikopylov, line 259: I'd prefer to return `ReadResult<>` here as well

src/blob/index/core.rs
```diff
@@ -302,21 +302,30 @@ where
...
+        };
+        Ok(if let Some(header) = result {
+            if header.is_deleted() {
+                ReadResult::Deleted(header.created())
+            } else {
+                ReadResult::Found(header)
+            }
+        } else {
+            ReadResult::NotFound
+        })
```
ikopylov, line 328: Merge conflicts with #209

src/blob/index/core.rs
```diff
@@ -302,21 +302,30 @@ where
         }
     }
 
-    async fn get_any(&self, key: &K) -> Result<Option<RecordHeader>> {
+    async fn get_any(&self, key: &K) -> Result<ReadResult<RecordHeader>> {
```
ikopylov, line 305: `get_all` above should also be updated to return `Result<ReadResult<Vec<RecordHeader>>>`

src/storage/core.rs
```diff
@@ -1212,3 +1297,37 @@ where
...
+    /// Data was found
+    Found(T),
+    /// Data was deleted, contains creation timestamp
+    Deleted(u64),
+    /// Data was not found
+    NotFound,
+}
+
+impl<T> ReadResult<T> {
+    pub(crate) fn is_meaningful(&self) -> bool {
```
ikopylov, line 1313: IMHO, the function name is not good. What do you think about  `is_presented` name?

Also add more functions for simplicity: 
1. `is_found`
2. `is_deleted`
3. `is_not_found`

src/storage/core.rs
```diff
@@ -304,7 +304,12 @@ where
         if self.try_create_active_blob().await.is_ok() {
             info!("Active blob was set during write operation");
         }
-        if !self.inner.config.allow_duplicates() && self.contains_with(key, meta.as_ref()).await? {
+        if !self.inner.config.allow_duplicates()
+            && matches!(
```
ikopylov, line 308: It is better to declare additional functions on `ReadResult<>` and use them instead of `match!` macro

src/storage/core.rs
```diff
@@ -441,26 +446,37 @@ where
                 future::ok(())
             })
             .await?;
+        let all_entries = filter_deleted(all_entries);
```
ikopylov, line 449: `read_all` should also return `ReadResult<>`

ikopylov: Please, recheck `all_entries` ordering. Looks like it is incorrect

src/storage/core.rs
```diff
@@ -781,27 +814,30 @@ where
...
         if let Some(active_blob) = &inner.active_blob {
-            if active_blob.read().await.contains(key, meta).await? {
-                return Ok(true);
+            let res = active_blob.read().await.contains(key, meta).await?;
+            if res.is_meaningful() {
+                return Ok(res);
             }
         }
         let blobs = inner.blobs.read().await;
         for blob in blobs.iter_possible_childs(key) {
```
ikopylov, line 833: `iter_possible_childs_rev`?

src/storage/core.rs
```diff
@@ -888,9 +924,11 @@ where
     /// Mark as deleted entries with matching key
     /// # Errors
     /// Fails after any disk IO errors.
-    pub async fn mark_all_as_deleted(&self, key: impl AsRef<K>) -> Result<u64> {
+    pub async fn mark_all_as_deleted(&self, key: impl AsRef<K>, force_write: bool) -> Result<u64> {
```
ikopylov, line 927: I don't like `mark_all_as_deleted` name for exported function. Maybe, it is better to rename it to `delete`?

Also, the new argument has a bad name too. I suggest inverting its meaning and changing its name to `only_if_presented`

src/storage/core.rs
```diff
@@ -933,6 +980,44 @@ where
     }
 }
 
+fn filter_deleted(all_entries: Vec<Entry>) -> Vec<Entry> {
+    let mut last_removal_by_key = HashMap::new();
+    let mut last_created_by_key = HashMap::new();
```
ikopylov, line 985: This function is only used from `read_all`, which returns records for one particular key. So you don't need a map for different keys. Additionally, you can enforce this by adding that key as the argument.

tests/tests.rs
```diff
@@ -899,8 +904,14 @@ async fn test_mark_as_deleted_single() {
...
     }
-    storage.mark_all_as_deleted(&delete_key).await.unwrap();
-    assert!(!storage.contains(delete_key).await.unwrap());
+    storage
+        .mark_all_as_deleted(&delete_key, false)
+        .await
+        .unwrap();
+    assert!(!matches!(
+        storage.contains(delete_key).await.unwrap(),
+        ReadResult::Found(_)
```
ikopylov, line 913: It should be `Deleted`

tests/tests.rs
```diff
@@ -7,7 +7,7 @@ use futures::{
     stream::{futures_unordered::FuturesUnordered, FuturesOrdered, StreamExt, TryStreamExt},
     TryFutureExt,
 };
-use pearl::{BloomProvider, Builder, Meta, Storage};
+use pearl::{BloomProvider, Builder, Meta, ReadResult, Storage};
```
ikopylov, line 10: We should have a test for `read_all` when deleted records presented in the middle

src/blob/core.rs
```diff
@@ -350,51 +358,61 @@ where
...
-            if let Some(header) = self.index.get_any(key).await.with_context(|| {
-                format!("index get any failed for blob: {:?}", self.name.to_path())
-            })? {
-                let entry = Entry::new(header, self.file.clone());
-                debug!("blob, get any entry, bloom true no meta, entry found");
-                Ok(Some(entry))
-            } else {
-                Ok(None)
-            }
+            Ok(self
```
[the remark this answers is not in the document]
idruzhitskiy: This way we can pass non-`Found` data freely through map, without using match

src/blob/core.rs
idruzhitskiy: Sure, once #209 is ready I'll merge it here

src/blob/core.rs
```diff
@@ -291,40 +301,50 @@ where
...
+    pub(crate) async fn read_all_entries(&self, key: &K) -> Result<ReadResult<Vec<Entry>>> {
         let headers = self.index.get_all(key).await?;
-        Ok(headers.map(|h| {
-            debug!("blob core read all {} headers", h.len());
-            Self::headers_to_entries(h, &self.file)
+        Ok(headers.map(|hs| {
+            debug_assert!(hs
+                .iter()
+                .zip(hs.iter().skip(1))
+                .all(|(x, y)| x.created() <= y.created()));
```
ikopylov, line 337: Looks like the condition should be reversed (`>=`)

src/blob/core.rs
```diff
@@ -350,51 +370,62 @@ where
         key: &K,
         meta: Option<&Meta>,
         check_filters: bool,
-    ) -> Result<Option<Entry>> {
-        debug!("blob get any entry {:?}, {:?}", key, meta);
+    ) -> Result<ReadResult<Entry>> {
+        println!("blob get any entry {:?}, {:?}", key, meta);
```
ikopylov, line 374: You should use logger here, not `println!`

src/blob/core.rs
```diff
@@ -350,51 +370,62 @@ where
...
+    async fn get_entry_with_meta(&self, key: &K, meta: &Meta) -> Result<ReadResult<Entry>> {
         let headers = self.index.get_all(key).await?;
-        if let Some(headers) = headers {
-            let entries = Self::headers_to_entries(headers, &self.file);
-            self.filter_entries(entries, meta).await
-        } else {
-            Ok(None)
-        }
+        headers
+            .try_map_async(|headers| async move {
```
ikopylov, line 401: I think `if` or `match` would be clearer here than `try_map_async`

src/blob/core.rs
```diff
@@ -350,51 +370,62 @@ where
...
+                let entries = Self::headers_to_entries(headers, &self.file);
+                self.filter_entries(entries, meta).await
+            })
+            .await
     }
 
-    async fn filter_entries(&self, entries: Vec<Entry>, meta: &Meta) -> Result<Option<Entry>> {
+    async fn filter_entries(&self, mut entries: Vec<Entry>, meta: &Meta) -> Result<Entry> {
+        entries.reverse();
```
ikopylov, line 409: I'm sorry, I was wrong: according by the latest changes in request #209, the actual order is from the newest entry to the oldest. So there is no need to reverse

src/blob/core.rs
```diff
@@ -350,51 +370,62 @@ where
...
         for mut entry in entries {
             if Some(meta) == entry.load_meta().await? {
-                return Ok(Some(entry));
+                return Ok(entry);
             }
         }
-        Ok(None)
+        Err(Error::new(ErrorKind::Other).into())
```
ikopylov, line 415: Why this is an error?

src/blob/index/core.rs
```diff
@@ -295,28 +298,51 @@ where
...
             State::OnDisk(findex) => findex.find_by_key(key).await,
-        }
+        }?;
+
+        Ok(if let Some(headers) = headers {
+            if let Some(header) = headers.last() {
+                if header.is_deleted() {
+                    ReadResult::Deleted(header.created())
+                } else {
+                    ReadResult::Found(headers)
```
ikopylov, line 312: Filtration logic from #209 should be here

src/blob/index/core.rs
```diff
@@ -295,28 +298,51 @@ where
...
-        match &self.inner {
+    async fn get_all(&self, key: &K) -> Result<ReadResult<Vec<RecordHeader>>> {
+        let headers = match &self.inner {
             State::InMemory(headers) => Ok(headers.get(key).cloned()),
             State::OnDisk(findex) => findex.find_by_key(key).await,
-        }
+        }?;
+
+        Ok(if let Some(headers) = headers {
+            if let Some(header) = headers.last() {
```
ikopylov, line 308: `first` should be used here (according to the recent changes in #209)

src/blob/index/core.rs
```diff
@@ -295,28 +298,51 @@ where
...
             }
             State::OnDisk(findex) => {
                 debug!("index get any on disk");
                 let header = findex.get_any(key).await?;
-                Ok(header)
+                header
             }
-        }
+        };
+        Ok(if let Some(header) = result {
```
ikopylov, line 337: Perhaps flat `match` statement would be clearer

src/storage/core.rs
```diff
@@ -1212,3 +1298,85 @@ where
         active + closed
     }
 }
+
+/// Result of read operations
+#[derive(Debug)]
+pub enum ReadResult<T> {
```
ikopylov, line 1304: It would be good to move `ReadResult<T>` into a separate file. Also `BlobRecordTimestamp` can be placed in the same file

src/storage/core.rs
```diff
@@ -413,54 +416,69 @@ where
...
-        }
         let blobs = safe.blobs.read().await;
         let entries_closed_blobs = blobs
-            .iter_possible_childs(key)
+            .iter_possible_childs_rev(key)
             .map(|b| b.1.data.read_all_entries(key))
             .collect::<FuturesUnordered<_>>();
         entries_closed_blobs
-            .try_filter_map(future::ok)
+            .try_filter_map(|r| async move { Ok(r.into_option()) })
```
ikopylov, line 429: Is `async` required here?

src/storage/core.rs
```diff
@@ -413,54 +416,69 @@ where
...
+            .as_ref()
+            .ok_or_else(Error::active_blob_not_set)?;
+        let read = active_blob.read().await.read_all_entries(key).await?;
+        if read.is_presented() {
+            let entries = read.unwrap();
+            debug!(
+                "storage core read all active blob entries {}",
+                entries.len()
+            );
+            all_entries.extend(entries);
```
ikopylov, line 451: According to the latest changes in #209 the newest records are at the head of vector, so the `active_blob` should be processed first

src/storage/core.rs
```diff
@@ -413,54 +416,69 @@ where
...
-                entries.len()
-            );
-            all_entries.extend(entries);
-        }
         let blobs = safe.blobs.read().await;
         let entries_closed_blobs = blobs
-            .iter_possible_childs(key)
+            .iter_possible_childs_rev(key)
             .map(|b| b.1.data.read_all_entries(key))
             .collect::<FuturesUnordered<_>>();
```
ikopylov, line 427: We lost ordering with `FuturesUnordered<_>`

src/storage/core.rs
ikopylov, line 429: With `r.into_option()` you lose the information about deletion

src/storage/core.rs
```diff
@@ -486,26 +504,43 @@ where
...
+                !h.is_presented()
+            } else {
+                true
+            }
+        });
+        task.next().await.unwrap_or(Ok(ReadResult::NotFound))
     }
 
-    async fn get_any_data(safe: &Safe<K>, key: &K, meta: Option<&Meta>) -> Result<Vec<u8>> {
+    async fn get_any_data(
```
ikopylov, line 539: Why this function needed? It is better to call directly `Self::get_data_last`

src/storage/core.rs
```diff
@@ -486,26 +504,43 @@ where
...
 
     #[allow(dead_code)]
-    async fn get_data_any(safe: &Safe<K>, key: &K, meta: Option<&Meta>) -> Result<Vec<u8>> {
+    async fn get_data_any(
+        safe: &Safe<K>,
+        key: &K,
+        meta: Option<&Meta>,
+    ) -> Result<ReadResult<Vec<u8>>> {
         let blobs = safe.blobs.read().await;
         let stream: FuturesUnordered<_> = blobs
```
ikopylov, line 524: Unfortunately, we should use `FuturesOrdered<_>`. Otherwise the deletion record can be skipped

src/storage/core.rs
idruzhitskiy: Yes

src/blob/index/core.rs
```diff
@@ -295,28 +298,51 @@ where
         }
     }
 
-    async fn get_all(&self, key: &K) -> Result<Option<Vec<RecordHeader>>> {
-        match &self.inner {
+    async fn get_all(&self, key: &K) -> Result<ReadResult<Vec<RecordHeader>>> {
```
ikopylov, line 301: My thoughts on the `get_all` function logic. At the `Storage` level, we need not only undeleted data, but also a `deleted` marker, which will allow to stop scanning old BLOBs. Unfortunately, that cannot be expressed with `ReadResult<>` type.
I think the `get_all` function can be split into 2 functions (on all levels: index, blob, storage):
1. `async fn get_all(&self, key: &K) -> Result<Vec<RecordHeader>>`: this version preserves the old behavior and returns only non-deleted records. If there are no non-deleted records, then it simply returns empty vector
2. `async fn get_all_with_deleted(&self, key: &K, go_beyond_deleted: bool) -> Result<Vec<RecordHeader>>`: this version returns all non-deleted records and then the record with `deleted` flag (if it was deleted). When `go_beyond_deleted == true` it also returns all records after the deleted one

Example for key `1`: v2, del, v1
1. get_all(1): `v2`
2. get_all_with_deleted(1, false): `v2, del`
3. get_all_with_deleted(1, true): `v2, del, v1`

In this way we can support all scenarios + this allow to stop scan on the deletion marker on `Storage` level

src/storage/core.rs
```diff
@@ -486,26 +504,43 @@ where
...
+            } else {
+                true
+            }
+        });
+        task.next().await.unwrap_or(Ok(ReadResult::NotFound))
     }
 
     #[allow(dead_code)]
-    async fn get_data_any(safe: &Safe<K>, key: &K, meta: Option<&Meta>) -> Result<Vec<u8>> {
+    async fn get_data_any(
```
ikopylov, line 518: I think we can remove this function. Previously it has a sense because it was supposed to be faster than getting the last record. Now, due to the deletion feature, it essentially becomes the same as `get_data_last`

src/blob/core.rs
idruzhitskiy: This function should be called only when header exists, so missing it is an error

src/blob/core.rs
idruzhitskiy: Latest should be the last, isn't it?...

src/blob/index/core.rs
idruzhitskiy: I suppose the order should be reversed? `v1, del, v2`

idruzhitskiy: Not sure why we need those. I've changed `read_all` on storage so it scans blobs and returns early if necessary. Is it necessary to search for records after deletion marker though?

ikopylov: For the user: we need a function that gets clear results without any special entries indicating deletion.
For internal needs (early stop on a removed record): we need a function that returns at least the first deletion marker.
The third option, which returns all records, ignoring deletion markers, is needed just in case (its implementation is trivial)

ikopylov:
Please, merge master. https://github.com/qoollo/pearl/pull/209 is already there

src/blob/core.rs
ikopylov: According to the #209 it is the first. I think this is a better order (most recent entries first)

src/blob/index/core.rs
idruzhitskiy: But isn't deletion supposed to be non-recoverable from the user's perspective?

ikopylov: Hm. Yes, that's a good argument.
Ok. Then I think the third option is not needed. We need just a 2 functions: 
1. `async fn get_all(&self, key: &K) -> Result<Vec<RecordHeader>>`
2. `async fn get_all_with_deletion_marker(&self, key: &K) -> Result<Vec<RecordHeader>>`

`get_all` can call `get_all_with_deletion_marker` where possible and remove the last item if it is a deletion marker

src/blob/core.rs
ikopylov: But this function filters out entries whose metadata is not equal to the passed `meta` value. The result can be empty if none of the entries has required values in the metadata

src/storage/core.rs
```diff
@@ -413,54 +416,80 @@ where
     /// Returns entries with matching key
     /// # Errors
     /// Fails after any disk IO errors.
-    pub async fn read_all(&self, key: impl AsRef<K>) -> Result<Vec<Entry>> {
+    pub async fn read_all(&self, key: impl AsRef<K>) -> Result<ReadResult<Vec<Entry>>> {
         let key = key.as_ref();
         let mut all_entries = Vec::new();
         let safe = self.inner.safe.read().await;
+        let blobs = safe.blobs.read().await;
```
ikopylov, line 423: Looks like it is better to acquire `read` lock not here, but below, just right before actual access to `blobs` collection

src/storage/core.rs
```diff
@@ -413,54 +416,80 @@ where
...
-            })
-            .await?;
-        debug!("storage core read all total {} entries", all_entries.len());
-        Ok(all_entries)
+            .collect::<FuturesOrdered<_>>();
+        while let Some(data) = futures.next().await {
+            let data = data?;
+            if data.is_found() {
+                all_entries.extend(data.unwrap());
+            } else if data.is_deleted() {
```
ikopylov, line 448: In the current implementation, it is impossible to stop on deletion if there is a record in the BLOB after the deletion marker, because `read_all_entries` do not return deletion marker in the collection.

Example:
- blob1: 
  - record_v1
- blob2 (active): 
  - record_v2
  - deletion_marker
  - record_v3

`read_all_entries` on blob2 will return `Found(vec![record_v3])`. 
Then `read_all_entries` on blob1 will return `Found(vec![record_v1])`.
The result will be: `Found(vec![record_v3, record_v1])`.
Deletion is incorrectly skipped, because there is no way to get the deletion marker from blob1 with the current implementation

src/storage/core.rs
```diff
@@ -885,12 +916,14 @@ where
...
-    /// Mark as deleted entries with matching key
+    /// Delete entries with matching key
     /// # Errors
     /// Fails after any disk IO errors.
-    pub async fn mark_all_as_deleted(&self, key: impl AsRef<K>) -> Result<u64> {
+    pub async fn delete(&self, key: impl AsRef<K>, only_if_presented: bool) -> Result<u64> {
         let mut total = 0;
-        total += self.mark_all_as_deleted_active(key.as_ref()).await?;
+        total += self
+            .mark_all_as_deleted_active(key.as_ref(), only_if_presented)
```
ikopylov, line 925: `write` lock on `safe` is acquired independently in `mark_all_as_deleted_active` and in `mark_all_as_deleted_closed`. This can lead to a situation where the active blob is replaced in between. I think, it is better to acquire that lock once in `delete` function

src/blob/index/core.rs
ikopylov: See the comment below for an example describing why we need the second function and why `ReadResult<>` does not work for the `get_all` case

src/storage/core.rs
idruzhitskiy: I see, thank you for clarification

src/blob/core.rs
```diff
@@ -294,40 +295,50 @@ where
...
     }
 
-    pub(crate) async fn mark_all_as_deleted(&mut self, key: &K) -> Result<bool> {
-        if self.index.get_any(key).await?.is_some() {
+    pub(crate) async fn mark_all_as_deleted(
+        &mut self,
+        key: &K,
+        only_if_presented: bool,
+    ) -> Result<bool> {
+        if !only_if_presented || !self.index.get_any(key).await?.is_found() {
```
ikopylov, line 341: Looks like an error, inversion on` get_any` is not needed: 
```suggestion
        if !only_if_presented || self.index.get_any(key).await?.is_found() {
```

src/storage/core.rs
```diff
@@ -415,100 +417,88 @@ where
...
+                        return Ok(data);
+                    }
                 }
                 Err(e) => debug!("read with optional meta active blob returned: {:#?}", e),
             }
         }
         Self::get_any_data(&safe, key, meta).await
     }
 
-    async fn get_data_last(safe: &Safe<K>, key: &K, meta: Option<&Meta>) -> Result<Bytes> {
```
ikopylov, line 465: Why it was removed? This is the important function that should be used to read data

src/storage/core.rs
```diff
@@ -415,100 +417,88 @@ where
...
-            .map(|blob| blob.data.read_any(key, meta, false))
-            .collect();
-        debug!("read with optional meta {} closed blobs", stream.len());
-        let mut task = stream.skip_while(Result::is_err);
-        task.next().await.ok_or_else(Error::not_found)?
-    }
-
-    #[allow(dead_code)]
-    async fn get_data_any(safe: &Safe<K>, key: &K, meta: Option<&Meta>) -> Result<Bytes> {
+    async fn get_any_data(
```
ikopylov, line 488: `get_any_data` is not needed and can be removed. You probably removed the wrong function

ikopylov:
@idruzhitskiy tests are failing

src/storage/core.rs
```diff
@@ -500,42 +500,42 @@ where
...
         debug!("read with optional meta {} closed blobs", stream.len());
         let mut task = stream.skip_while(Result::is_err);
-        let res = task.next().await;
-        res.unwrap_or(Ok(ReadResult::NotFound))
-    }
 
-    /// Stop blob updater and release lock file
-    /// # Errors
-    /// Fails because of any IO errors
-    pub async fn close(self) -> Result<()> {
```
ikopylov, line 517: Why `close` was removed?

src/blob/core.rs
```diff
@@ -322,8 +322,11 @@ where
     }
 
     #[inline]
-    pub(crate) async fn read_all_entries(&self, key: &K) -> Result<ReadResult<Vec<Entry>>> {
-        let headers = self.index.get_all(key).await?;
+    pub(crate) async fn read_all_entries_with_deletion_marker(
```
ikopylov, line 325: I think it would be good to have `read_all_entries` on the Blob level too

src/blob/core.rs
```diff
@@ -322,8 +322,11 @@ where
     }
 
     #[inline]
-    pub(crate) async fn read_all_entries(&self, key: &K) -> Result<ReadResult<Vec<Entry>>> {
```
ikopylov: I think it would be good to have `read_all_entries` on the Blob level too (in addition to `read_all_entries_with_deletion_marker`)

src/blob/index/core.rs
```diff
@@ -292,6 +292,20 @@ where
...
+                    hs.truncate(hs.len() - 1);
+                }
+                hs
+            } else {
+                hs
+            }
+        }))
+    }
+
+    async fn get_all_with_deletion_marker(&self, key: &K) -> Result<ReadResult<Vec<RecordHeader>>> {
```
ikopylov, line 308: Returning `ReadResult<>` only makes things harder to process on upper level. I think it is better to return just `Result<Vec<RecordHeader>>`

src/blob/core.rs
```diff
@@ -322,8 +322,11 @@ where
...
-        let headers = self.index.get_all(key).await?;
+    pub(crate) async fn read_all_entries_with_deletion_marker(
+        &self,
+        key: &K,
+    ) -> Result<ReadResult<Vec<Entry>>> {
```
ikopylov, line 328: The same here: better simplify and return `Result<Vec<Entry>>`

src/blob/core.rs
```diff
@@ -294,40 +295,53 @@ where
...
+                debug!("blob read any entry loaded bytes: {}", buf.len());
+                Ok(ReadResult::Found(buf))
+            }
+            ReadResult::Deleted(ts) => Ok(ReadResult::Deleted(ts)),
+            ReadResult::NotFound => Ok(ReadResult::NotFound),
+        }
     }
 
     #[inline]
-    pub(crate) async fn read_all_entries(&self, key: &K) -> Result<Option<Vec<Entry>>> {
```
ikopylov, line 321: I think it would be good to have `read_all_entries` on the Blob level too (in addition to `read_all_entries_with_deletion_marker`)

src/blob/core.rs
```diff
@@ -357,35 +371,42 @@ where
...
+                    let entry = Entry::new(header, self.file.clone());
+                    debug!("blob, get any entry, bloom true no meta, entry found");
+                    entry
+                }))
         }
     }
 
-    async fn get_entry_with_meta(&self, key: &K, meta: &Meta) -> Result<Option<Entry>> {
+    async fn get_entry_with_meta(&self, key: &K, meta: &Meta) -> Result<ReadResult<Entry>> {
         let headers = self.index.get_all(key).await?;
```
ikopylov, line 400: `get_all_with_deletion_marker` should be used here. If no record passed filtration by `meta` and the last record is a deletion marker, then `ReadResult::Deleted()` should be returned

src/storage/core.rs
```diff
@@ -415,54 +417,97 @@ where
     /// Returns entries with matching key
     /// # Errors
     /// Fails after any disk IO errors.
-    pub async fn read_all(&self, key: impl AsRef<K>) -> Result<Vec<Entry>> {
+    pub async fn read_all(&self, key: impl AsRef<K>) -> Result<ReadResult<Vec<Entry>>> {
```
ikopylov, line 420: It is better to return just `Result<Vec<Entry>>`

ikopylov, line 420: I propose to add `read_all_with_deletion_marker` here too. `read_all` can call it and remove the last item if it is a deletion marker

src/storage/core.rs
```diff
@@ -415,54 +417,97 @@ where
...
             .active_blob
             .as_ref()
             .ok_or_else(Error::active_blob_not_set)?;
-        if let Some(entries) = active_blob.read().await.read_all_entries(key).await? {
+        let read = active_blob
+            .read()
+            .await
+            .read_all_entries_with_deletion_marker(key)
+            .await?;
+        if read.is_found() {
```
ikopylov, line 433: Rejecting `ReadResult<>` in `read_all_entries_with_deletion_marker` return type will simplify the code, allowing you to remove this `if` statement.

src/blob/index/core.rs
idruzhitskiy: Indeed, thank you

ikopylov:
We still need a test for `read_all_with_deletion_marker` on Storage level when deleted records presented in the middle

src/storage/core.rs
```diff
@@ -416,53 +418,93 @@ where
...
-        debug!("storage core read all total {} entries", all_entries.len());
+        let mut futures = blobs
+            .iter_possible_childs_rev(key)
+            .map(|b| b.1.data.read_all_entries_with_deletion_marker(key))
+            .collect::<FuturesOrdered<_>>();
+        while let Some(data) = futures.next().await {
+            let entries = data?;
+            if let Some(e) = entries.last() {
+                if e.header().is_deleted() {
+                    all_entries.extend(entries);
```
ikopylov, line 465: `extend` is performed regardless of whether the deletion marker is present or not. So I think it would be good to move this line up to the position right after the `entries` are assigned (`let entries = data?;`)

idruzhitskiy: Extend requires ownership in this case, and if we add it earlier we need to check last record in `all_entries` and remember if any was added from `entries` this time, which will require new flag. I think this way there are less lines of code involved

src/storage/core.rs
```diff
@@ -416,53 +418,93 @@ where
...
+    /// # Errors
+    /// Fails after any disk IO errors.
+    pub async fn read_all_with_deletion_marker(&self, key: impl AsRef<K>) -> Result<Vec<Entry>> {
         let key = key.as_ref();
         let mut all_entries = Vec::new();
         let safe = self.inner.safe.read().await;
         let active_blob = safe
             .active_blob
             .as_ref()
             .ok_or_else(Error::active_blob_not_set)?;
```
ikopylov, line 440: This is incorrect. We should skip `active_blob` if it is not set

tests/tests.rs
```diff
@@ -1089,3 +1089,24 @@ async fn test_read_all_with_deletion_marker_delete_middle() -> Result<()> {
...
+async fn test_read_all_with_deletion_marker_delete_middle_different_blobs() -> Result<()> {
+    let path = common::init("new");
+    let storage = common::default_test_storage_in(path).await.unwrap();
+    let key: KeyTest = vec![0].into();
+    let data: Bytes = "test data string".repeat(16).as_bytes().to_vec().into();
+    storage.write(&key, data.clone()).await?;
+    storage.try_close_active_blob().await?;
+    storage.delete(&key, true).await?;
+    storage.try_close_active_blob().await?;
+    storage.write(&key, data.clone()).await?;
```
ikopylov, line 1103: Add `storage.try_close_active_blob().await?;` after this `write` too. This will allow to verify correct working, when active blob is closed

ikopylov (approved):
LGTM

PR #211, opened 2022-09-29
190 file ops update

### LINKED ISSUE #190: Bad writing data method used in `File::write_at_sync`
It uses `write_at` which does not guarantee that all passed data will be written. Instead `write_all_at` should be used (https://doc.rust-lang.org/std/os/unix/fs/trait.FileExt.html#method.write_all_at)

https://github.com/qoollo/pearl/blob/master/src/blob/file.rs#L96

idruzhitskiy:
Closes #190

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ Pearl changelog
 
 
 #### Changed
-
+- Change file operations to work on full buffers (#190)
 
 #### Fixed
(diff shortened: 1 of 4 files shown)


ikopylov (approved):
LGTM

ikopylov:
@idruzhitskiy please, resolve merge conflicts

PR #212, opened 2022-10-19
192 large data optimization

### LINKED ISSUE #189: Improve `Record` serialization
`Record:to_raw` is inefficient, because it creates multiple vectors. It should create single vector (ideally with preallocated capacity) and then serialize all data into that single vector.

https://github.com/qoollo/pearl/blob/master/src/record.rs#L111

### LINKED ISSUE #192: Avoid copying data during serialization of `Record`
Currently, the whole data copied to the final vector here: https://github.com/qoollo/pearl/blob/master/src/record.rs#L116
This becomes a problem, when data is large (~1Gb). 
We can try to avoid this copy and pass data as a separate parameter to `write_append` here: https://github.com/qoollo/pearl/blob/master/src/blob/core.rs#L240

There is `write_all_vectored` function (https://doc.rust-lang.org/std/io/trait.Write.html#method.write_all_vectored) to write 2 buffers in a single call, but it is nightly. Still not copying should provide a significant improvement for large data even with two write calls.

If the data is small (<4Kb) we can use the current approach with single buffer. For large data 2 calls is better

idruzhitskiy:
Closes #192 
Closes #189

### DIFF
--- a/src/blob/core.rs
+++ b/src/blob/core.rs
@@ -1,7 +1,7 @@
 use std::time::SystemTime;
 
 use async_std::sync::RwLockUpgradableReadGuard;
-use bytes::Bytes;
+use bytes::{BufMut, Bytes, BytesMut};
 use tokio::time::Instant;
 
 use crate::error::ValidationErrorKind;
@@ -71,9 +71,11 @@ where
     }
 
     async fn write_header(&mut self) -> Result<()> {
-        let buf = serialize(&self.header)?;
-        self.file.write_append(&buf).await?;
-        self.current_offset = buf.len() as u64;
+        let mut buf = BytesMut::new();
+        serialize_into((&mut buf).writer(), &self.header)?;
+        let len = buf.len() as u64;
+        self.file.write_append_all(buf.freeze()).await?;
+        self.current_offset = len;
         Ok(())
     }
 
--- a/src/blob/core.rs
+++ b/src/blob/core.rs
@@ -94,10 +94,12 @@ where
                 .await
                 .with_context(|| format!("blob file dump failed: {:?}", self.name.to_path()))?;
 
-            self.index
-                .dump(self.file_size())
-                .await
-                .with_context(|| format!("index file dump failed, associated blob file: {:?}", self.name.to_path()))
+            self.index.dump(self.file_size()).await.with_context(|| {
+                format!(
+                    "index file dump failed, associated blob file: {:?}",
+                    self.name.to_path()
+                )
+            })
         }
     }
 
@@ -146,7 +148,10 @@ where
                 if let Some(io_error) = error.downcast_ref::<IOError>() {
                     match io_error.kind() {
                         IOErrorKind::PermissionDenied | IOErrorKind::Other => {
-                            warn!("index for file '{:?}' cannot be regenerated due to an error: {}", path, io_error);
+                            warn!(
+                                "index for file '{:?}' cannot be regenerated due to an error: {}",
+                                path, io_error
+                            );
                             return Err(error);
                         }
                         _ => {}
@@ -205,10 +210,12 @@ where
             return Ok(());
         }
         debug!("index file missed");
-        let raw_r = self
-            .raw_records()
-            .await
-            .with_context(|| format!("failed to read raw records from blob {:?}", self.name.to_path()))?;
+        let raw_r = self.raw_records().await.with_context(|| {
+            format!(
+                "failed to read raw records from blob {:?}",
+                self.name.to_path()
+            )
+        })?;
         debug!("raw records loaded");
         if let Some(headers) = raw_r.load().await.with_context(|| {
             format!(
@@ -234,19 +241,7 @@ where
         let mut offset = self.current_offset.lock().await;
         debug!("blob write record offset: {}", *offset);
         record.set_offset(*offset)?;
-        let buf = record.to_raw()?;
-        let bytes_written = self
-            .file
-            .write_append(&buf)
-            .await
-            .map_err(|e| -> anyhow::Error {
-                match e.kind() {
-                    kind if kind == IOErrorKind::Other || kind == IOErrorKind::NotFound => {
-                        Error::file_unavailable(kind).into()
-                    }
-                    _ => e.into(),
-                }
-            })? as u64;
+        let bytes_written = record.write_to_file(&self.file).await?;
         self.index.push(record.header().clone())?;
         *offset += bytes_written;
         Ok(())
@@ -267,7 +262,14 @@ where
         let buf = entry
             .load()
             .await
-            .with_context(|| format!("failed to read key {:?} with meta {:?} from blob {:?}", key, meta, self.name.to_path()))?
+            .with_context(|| {
+                format!(
+                    "failed to read key {:?} with meta {:?} from blob {:?}",
+                    key,
+                    meta,
+                    self.name.to_path()
+                )
+            })?
             .into_data();
         debug!("blob read any entry loaded bytes: {}", buf.len());
         Ok(buf)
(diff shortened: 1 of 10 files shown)


src/blob/file.rs
```diff
@@ -50,12 +50,13 @@ impl File {
         self.size.load(ORD)
     }
 
-    pub(crate) async fn write_append(&self, buf: &[u8]) -> IOResult<()> {
+    pub(crate) async fn write_append(&self, buf: &[u8]) -> IOResult<usize> {
```
ikopylov, line 53: Perhaps `write_at` should be updated in the same way. What do you think?

src/record.rs
```diff
@@ -106,15 +111,46 @@ impl Record {
...
+        let mut buf = self.create_header_meta_buffer()?;
         buf.extend(&self.data);
-        Ok(buf)
+        Self::process_file_result(file.write_append(&buf).await)
+    }
+
+    async fn write_double_pass(&self, file: &File) -> Result<u64> {
+        let buf = self.create_header_meta_buffer()?;
+        Self::process_file_result(
+            file.write_append(&buf)
```
ikopylov, line 139: `file.write_append` call gives overhead due to usage of `spawn_blocking`. It is better to add an additional function that takes 2 buffers and write them. This will additionally allow to use `write_all_vectored` (https://doc.rust-lang.org/std/io/trait.Write.html#method.write_all_vectored) in future

src/blob/file.rs
idruzhitskiy: Currently there is no reason, I think it should be updated when needed

src/blob/file.rs
```diff
@@ -69,6 +69,44 @@ impl File {
         self.write_at_aio(offset, buf, ioring).await
     }
 
+    pub(crate) async fn write_append_buffers(&self, buffers: Vec<&[u8]>) -> IOResult<usize> {
```
ikopylov, line 72: `Vec<>` implies memory allocation. It is better to pass slice of slices. Another possibility: we know that we only need a case for 2 buffers, so we can pass 2 buffers in separate arguments.

src/blob/file.rs
```diff
@@ -69,6 +69,44 @@ impl File {
...
+            self.write_at_aio(offset, buf, ioring).await?;
+            offset = offset + buf.len() as u64;
+        }
+        Ok(())
+    }
+
+    async fn write_append_buffers_sync(&self, buffers: Vec<&[u8]>, len: usize) -> IOResult<()> {
+        let mut offset = self.size.fetch_add(len as u64, Ordering::SeqCst);
+        let file = self.no_lock_fd.clone();
+        let buffers: Vec<_> = buffers.into_iter().map(|v| v.to_vec()).collect();
```
ikopylov, line 99: The idea was to avoid copying data as much as possible, but here you're converting a slice to a vector, which implies copying

ikopylov: As I see, `write_at_sync` is also copy data into vector. This also should be changed.
If there is some issue with passing of ownership, then probably we can use `Bytes` (can be done in #213)

idruzhitskiy: It can'be passed without copy indeed, since we need to pass ownership to spawn_blocking. Should I merge this branch into #213 or wait until it's merged into master?

ikopylov: I think, it is better to wait until this branch will be merged

ikopylov: Or we can do it in reverse: merge #213 to master, because it is ready, and then merge master to the current branch and update to use `Bytes` here

ikopylov: I've merged #213

idruzhitskiy: Done

src/blob/file.rs
```diff
@@ -56,33 +56,94 @@ impl File {
...
+        first_buf: Bytes,
+        second_buf: Bytes,
+    ) -> IOResult<()> {
+        let first_len = first_buf.len() as u64;
+        let second_len = second_buf.len() as u64;
+        let len = first_len + second_len;
+        let mut offset = self.size.fetch_add(len, Ordering::SeqCst);
+        let file = self.no_lock_fd.clone();
+        if len <= MAX_SYNC_OPERATION_SIZE as u64 {
+            Self::inplace_sync_call(move || {
```
ikopylov, line 121: I think, that for 2 buffers case we should always use `background_sync_call`. The problem is that `MAX_SYNC_OPERATION_SIZE` was chosen for single operation and cannot be safely applied to the case with 2 blocking operations in row

src/blob/core.rs
```diff
@@ -71,9 +71,11 @@ where
     }
 
     async fn write_header(&mut self) -> Result<()> {
-        let buf = serialize(&self.header)?;
-        self.file.write_append(&buf).await?;
-        self.current_offset = buf.len() as u64;
+        let mut buf = BytesMut::new();
```
ikopylov, line 74: It would be good to estimate expected size and create buffer with capacity. Bincode has `serialized_size` function for that

src/blob/index/bptree/core.rs
```diff
@@ -62,10 +62,11 @@ where
         let file = File::create(path, ioring)
             .await
             .with_context(|| format!("file open failed {:?}", path))?;
-        file.write_append(&buf).await?;
+        file.write_append_all(buf.freeze()).await?;
         header.set_written(true);
-        let serialized_header = serialize(&header)?;
-        file.write_at(0, &serialized_header).await?;
+        let mut serialized_header = BytesMut::new();
```
ikopylov, line 67: Same here: It would be good to estimate expected size and create buffer with capacity. Bincode has `serialized_size` function for that

src/blob/index/simple.rs
```diff
@@ -79,10 +79,11 @@ where
         let file = File::create(path, ioring)
             .await
             .with_context(|| format!("file open failed {:?}", path))?;
-        file.write_append(&buf).await?;
+        file.write_append_all(buf.freeze()).await?;
         header.set_written(true);
-        let serialized_header = serialize(&header)?;
-        file.write_at(0, &serialized_header).await?;
+        let mut serialized_header = BytesMut::new();
```
ikopylov, line 84: And here: It would be good to estimate expected size and create buffer with capacity. Bincode has `serialized_size` function for that

src/blob/core.rs
ikopylov: You can add `serialized_size` function to BLOB header struct like it is done for record header

src/blob/index/bptree/core.rs
ikopylov: You can add `serialized_size` function to Index header struct like it is done for record header

src/blob/header.rs
```diff
@@ -60,4 +60,9 @@ impl Header {
         }
         Ok(())
     }
+
+    #[inline]
+    pub(crate) fn serialized_size(&self) -> bincode::Result<u64> {
+        serialized_size(&self)
```
ikopylov, line 66: It is hard to imagine, that this can fail. Please, use `expect` and convert result to `usize`

src/record.rs
```diff
@@ -1,7 +1,12 @@
-use crate::{error::ValidationErrorKind, prelude::*};
+use std::io::Cursor;
+
+use futures::TryFutureExt;
+
+use crate::{blob::File, error::ValidationErrorKind, prelude::*};
 
 pub(crate) const RECORD_MAGIC_BYTE: u64 = 0xacdc_bcde;
 const DELETE_FLAG: u8 = 0x01;
+const MAX_SINGLE_PASS_DATA_SIZE: usize = 4 * 1024;
```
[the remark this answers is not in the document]
ikopylov: Constant verified by tests

PR #213, opened 2022-10-21
191 bytes arguments

### LINKED ISSUE #191: Investigate the possibility of receiveing data as a slice
Currently, data being passed as a vector, which requires data to be clonned on Bob's side. Probably, we can receive data as a slice `&[u8]`

https://github.com/qoollo/pearl/blob/master/src/storage/core.rs#L249

idruzhitskiy:
Closes #191

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Pearl changelog
 #### Changed
 - File name added to error messages in Blob struct (#183)
 - Update rust edition (#175)
+- Use bytes for arguments (#191)
 
 #### Fixed
 - Fix docs (#177)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Pearl changelog
 
 
 #### Changed
+- Use bytes for arguments (#191)
 
 
 #### Fixed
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -41,6 +41,7 @@ Pearl changelog
 #### Changed
 - File name added to error messages in Blob struct (#183)
 - Update rust edition (#175)
+- Use bytes for arguments (#191)
```
ikopylov, line 44: Move this line to `[Unreleased]` section

Cargo.toml
```diff
@@ -44,6 +44,7 @@ tokio-stream = "0.1"
 nix = "0.25.0"
 libc = "0.2"
 static_assertions = "1.1.0"
+bytes = "1.2.1"
```
ikopylov, line 47: `1.2` is better. This will allow updates that do not break backwards compatibility

tests/tests.rs
```diff
@@ -416,7 +420,7 @@ async fn test_read_with() {
     debug!("read finished");
     // data_read - last record, data_read_with - first record with "1.0" meta
     assert_ne!(data_read_with, data_read);
-    assert_eq!(data_read_with, data1);
+    assert_eq!(data_read_with.to_vec(), data1);
```
ikopylov, line 423: Same here

ikopylov (approved):
LGTM

PR #214, opened 2022-10-30
464 corrupted blobs count

Boneyan:
Corrupted blobs count calculation

### DIFF
--- a/src/storage/core.rs
+++ b/src/storage/core.rs
@@ -65,7 +65,7 @@ where
     pub(crate) blobs: Arc<RwLock<HierarchicalFilters<K, CombinedFilter<K>, Blob<K>>>>,
 }
 
-async fn work_dir_content(wd: &Path) -> Result<(Vec<DirEntry>, bool)> {
+async fn work_dir_content(wd: &Path) -> Result<Option<Vec<DirEntry>>> {
     let mut files = Vec::new();
     let mut dir = read_dir(wd).await?;
     while let Some(file) = dir.next_entry().await.transpose() {
@@ -80,10 +80,10 @@ async fn work_dir_content(wd: &Path) -> Result<(Vec<DirEntry>, bool)> {
         .any(|name| name.ends_with(BLOB_FILE_EXTENSION))
     {
         debug!("working dir contains files, try init existing");
-        (files, true)
+        Some(files)
     } else {
         debug!("working dir is uninitialized, starting empty storage");
-        (files, false)
+        None
     };
     Ok(content)
 }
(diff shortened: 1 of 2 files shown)


idruzhitskiy (approved):

src/storage/core.rs
```diff
@@ -710,13 +721,37 @@ where
...
             }
         }
-        Ok(blobs)
+        Ok((blobs, corrupted))
+    }
+
+    async fn count_old_corrupted_blobs(files: &[DirEntry], config: &Config) -> usize {
+        let dir_content = files.iter().map(DirEntry::path);
+        let dirs = dir_content.filter(|path| 
+            path.is_dir() && path.ends_with(config.corrupted_dir_name()));
```
ikopylov, line 737: Use `file_name` instead of `ends_with`. `ends_with` will not work, for example, if the path ends with `/`

src/storage/core.rs
```diff
@@ -64,7 +65,7 @@ where
     pub(crate) blobs: Arc<RwLock<HierarchicalFilters<K, CombinedFilter<K>, Blob<K>>>>,
 }
 
-async fn work_dir_content(wd: &Path) -> Result<Option<Vec<DirEntry>>> {
+async fn work_dir_content(wd: &Path) -> Result<(Vec<DirEntry>, bool)> {
```
ikopylov, line 68: I don't like this change. It is better to rewrite `count_old_corrupted_blobs` to not receive `[DirEntry]` as parameter

src/storage/core.rs
```diff
@@ -712,13 +729,39 @@ where
...
                         return Err(e.context(msg));
                     }
                 }
             }
         }
-        Ok(blobs)
+        Ok((blobs, corrupted))
+    }
+
+    async fn count_old_corrupted_blobs(files: &[DirEntry], config: &Config) -> usize {
```
ikopylov, line 742: I think it is better to take the path to the pearl directory rather than all of its content, because we only need one directory inside of it. You can combine pearl directory with `config.corrupted_dir_name()` and get an expected path. Then you can easily create `Path` from it, check its exists and then call `read_dir`

src/storage/core.rs
```diff
@@ -548,6 +552,11 @@ where
         }
     }
 
+    /// `blob_count` returns exact number of corrupted blobs.
+    pub async fn corrupted_blobs_count(&self) -> usize {
```
ikopylov, line 556: `async` is not needed here

src/storage/core.rs
```diff
@@ -712,13 +726,43 @@ where
...
+    async fn count_old_corrupted_blobs(config: &Config) -> usize {
+        let mut corrupted = 0;
+
+        if let Some(work_dir_path) = config.work_dir() {
+            let mut corrupted_dir_path = work_dir_path.to_path_buf();
+            corrupted_dir_path.push(config.corrupted_dir_name());
+            if corrupted_dir_path.exists() {
+                let dir = read_dir(&corrupted_dir_path).await;
+                if dir.is_err() {
+                    warn!("can't read corrupted blobl dir: {}", corrupted_dir_path.display());
```
ikopylov, line 748: Typo: `"can't read corrupted blob dir: {}"`
Also, add error description to the message

ikopylov (approved):
LGTM

PR #219, opened 2022-11-18
218 filters inner locks

### LINKED ISSUE #218: Make filter update not require mutable reference
Related to https://github.com/qoollo/bob/issues/596

idruzhitskiy:
Closes #218

### DIFF
--- a/src/filter/atomic_bitvec.rs
+++ b/src/filter/atomic_bitvec.rs
@@ -1,5 +1,6 @@
 use std::sync::atomic::*;
 
+/// Bit vector with atomic operations on its bits
 pub(crate) struct AtomicBitVec {
     data: Vec<AtomicU64>,
     bits_count: usize
@@ -51,8 +52,10 @@ impl AtomicBitVec {
     }
 
 
-    #[inline]
-    pub(crate) fn get(&self, index: usize, ordering: Ordering) -> bool {
+    /// Reads bit value at 'index' offset with specific ordering
+    /// Ordering possible values are [`Ordering::SeqCst`], [`Ordering::Acquire`] and [`Ordering::Relaxed`]
+    #[inline(always)]
+    pub(crate) fn get_ord(&self, index: usize, ordering: Ordering) -> bool {
         debug_assert!(index < self.bits_count);
 
         let (offset, mask) = Self::offset_and_mask(index);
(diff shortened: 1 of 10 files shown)


idruzhitskiy:
After tests this doesn't seem to improve much on local machine, so maybe it does not worth it

src/filter/bloom.rs
```diff
@@ -3,12 +3,13 @@ use ahash::AHasher;
...
 use bitvec::prelude::*;
 use std::hash::Hasher;
+use std::sync::RwLock;
 
 // All usizes in structures are serialized as u64 in binary
 #[derive(Clone)]
 /// Bloom filter
 pub struct Bloom {
-    inner: Option<BitVec<u64, Lsb0>>,
+    inner: Option<Arc<RwLock<BitVec<u64, Lsb0>>>>,
```
ikopylov, line 12: Looks like `BitVec` is actually a thread safe: https://docs.rs/bitvec/latest/bitvec/#feature-flags
This should be checked additionally.

src/filter/range.rs
```diff
@@ -7,18 +8,22 @@ where
...
+    min: Arc<RwLock<K>>,
     #[serde(serialize_with = "serialize_key", deserialize_with = "deserialize_key")]
-    max: K,
-    initialized: bool,
+    max: Arc<RwLock<K>>,
+    #[serde(
+        serialize_with = "serialize_initialized",
+        deserialize_with = "deserialize_initialized"
+    )]
+    initialized: Arc<RwLock<bool>>,
```
ikopylov, line 18: Three `RwLock<>` is too much. You should split `RangeFilter<>` into 2 structs:
1. `RangeFilterSafe<>` that assumes to work under protection (basically it is the copy of the original `RangeFilter<>`)
2. `RangeFilter<>` that contains only one field: `RwLock<RangeFilterSafe<>>` and redirects all operations to `RangeFilterSafe<>`

src/filter/bloom.rs
ikopylov: According to the source code it uses atomic `fetch_or` and `fetch_and` if the stored data type is `AtomicU...`. But at the same time I did not find a function in public API that allows mutation in a thread safe way. Please, recheck that.
If there are actually no thread safe functions in the API, then I think your initial idea to implement `BitVec` as `Vec<AtomicU64>` is good. There is also `atomic_bitvec` crate, but it don't have all the functions we need (conversion to and from vec, merging)

ikopylov:
There should also be PR on Bob side that uses this improvement

src/filter/atomic_bitvec.rs
```diff
@@ -0,0 +1,355 @@
...
+}
+
+#[derive(Debug)]
+pub(crate) enum AtomicBitVecError {
+    BitsCountMismatch,
+    DataLengthLessThanRequired
+}
+
+/// Contains helper functions for bit access
+pub(crate) struct OffsetAndMaskCalculator<TItem: Sized> {
```
idruzhitskiy, line 16: Is generic really needed? The only place it's being used is with `u8`, why not make it `ByteOffsetAndMaskCalculator`?

src/filter/atomic_bitvec.rs
```diff
@@ -0,0 +1,355 @@
...
+
+    #[inline(always)]
+    pub(crate) const fn get_bit(data: u8, mask: u8) -> bool {
+        (data & mask) != 0
+    }
+}
+
+
+impl AtomicBitVec {
+    const ITEM_BYTES_SIZE: usize = 8;
```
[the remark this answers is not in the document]
ikopylov: I changed `ITEM_BITS_SIZE` to be calculated from the `ITEM_BYTES_SIZE` value: 
```
const ITEM_BITS_SIZE: usize = Self::ITEM_BYTES_SIZE * 8;
```

src/filter/atomic_bitvec.rs
```diff
@@ -0,0 +1,355 @@
...
+    const ITEM_BITS_SIZE: usize = 64;
+
+    #[inline(always)]
```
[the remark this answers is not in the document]
ikopylov: This function is used on the hot paths. Normally, RUST should inline it itself. Directive added just in case

src/filter/atomic_bitvec.rs
ikopylov: Removed generic parameter and added `_u8` suffix to the function names

idruzhitskiy:
LGTM

PR #220, opened 2022-11-22
188 serialize outside lock

### LINKED ISSUE #188: Place lock on `current_offset` right before actual write to file in `Blob`
Lock taken too early. First, serialization should be performed and only after that the lock should be taken.
https://github.com/qoollo/pearl/blob/master/src/blob/core.rs#L234

This will require `offset` patching, but should improve performance

### LINKED ISSUE #199: Rename `Blob::read_any` to `Blob::read_last`
In the current implementation, `Blob::read_any` actually reads the last record and we strongly depend on this behavior.
https://github.com/qoollo/pearl/blob/master/src/blob/core.rs#L255

### LINKED ISSUE #206: Record header checksum should be calculated without additional header srialization
Currently, header checksum calculation calls serialization: https://github.com/qoollo/pearl/blob/master/src/record.rs#L240
We can improve the things by calculating the header checksum directly in `to_raw` function: in this function serialization can be called with `0` in `header_checksum` field, then `CRC` can be calculated on the serialized data, then the result can be written directly into serialized bytes vector by hand.

idruzhitskiy:
Closes #188 
Closes #206 
Closes #199


src/record.rs
```diff
@@ -230,6 +279,11 @@ impl Header {
         bincode::serialized_size(&self).expect("calc record serialized size")
     }
 
+    pub(crate) fn set_offset_checksum(&mut self, blob_offset: u64, header_checksum: u32) {
+        self.blob_offset = blob_offset;
+        self.header_checksum = header_checksum;
```
ikopylov, line 284: Can you add `debug_assert` to verify, that new checksum equals to `self.crc32()`

src/record.rs
```diff
@@ -106,15 +152,19 @@ impl Record {
         &self.header
     }
 
-    /// # Description
-    /// Serialize record to bytes
-    pub fn to_raw(&self) -> bincode::Result<Vec<u8>> {
+    pub(crate) fn to_partially_serialized(&self) -> Result<PartiallySerializedRecord> {
```
ikopylov, line 155: This logic should be slightly modified after https://github.com/qoollo/pearl/pull/212 will be merged

src/record.rs
```diff
@@ -80,6 +80,52 @@ impl Meta {
     }
 }
 
+pub(crate) struct PartiallySerializedRecord {
+    buf: Vec<u8>,
+    offset_pos: usize,
+    header_checksum_pos: usize,
```
ikopylov, line 86: `offset_pos` and `header_checksum_pos` are constants, so no need to store them as a part of `PartiallySerializedRecord`

src/record.rs
```diff
@@ -80,6 +80,52 @@ impl Meta {
...
+    header_start: usize,
```
ikopylov, line 87: As I see, `header_start` is always `0`, so no need to store it here

src/record.rs
```diff
@@ -106,15 +152,19 @@ impl Record {
...
-        trace!("raw header: len: {}", buf.len());
+        let header_len = buf.len();
         let raw_meta = self.meta.to_raw()?;
         buf.extend(&raw_meta);
         buf.extend(&self.data);
-        Ok(buf)
+        Ok(PartiallySerializedRecord::new(
+            buf,
+            header_len - 24,
+            header_len - 4,
```
ikopylov, line 164: Add functions to `Header` that evaluate offsets:
`fn get_blob_offset_offset(header_len: usize) -> usize`
`fn get_data_checksum_offset(header_len: usize) -> usize`

These function can be used in `into_serialized_with_header_checksum`

ikopylov: Also it is super critical to write tests, that verifies correctness of that functions. 
Test logic: serialize header, update data at the calculated offsets, deserialize header, verify that value was correctly updated

src/record.rs
```diff
@@ -80,6 +80,52 @@ impl Meta {
     }
 }
 
+pub(crate) struct PartiallySerializedRecord {
+    buf: Vec<u8>,
```
ikopylov, line 84: Probably, this should be `BytesMut`

src/record.rs
```diff
@@ -80,6 +80,52 @@ impl Meta {
...
+    ) -> Result<(Vec<u8>, u32)> {
+        let Self {
+            mut buf,
+            offset_pos,
+            header_checksum_pos,
+            header_start,
+            header_len,
+        } = self;
+        let offset_slice = &mut buf[offset_pos..];
+        bincode::serialize_into(offset_slice, &blob_offset)?;
```
ikopylov, line 120: I suggest to use `offset_slice.copy_from_slice(&blob_offset.to_le_bytes());`
```suggestion
        offset_slice.copy_from_slice(&blob_offset.to_le_bytes());
```

src/record.rs
```diff
@@ -80,6 +80,52 @@ impl Meta {
...
+        let header_slice = &buf[header_start..(header_start + header_len)];
+        let checksum = CRC32C.checksum(header_slice);
+        let checksum_slice = &mut buf[header_checksum_pos..];
+        bincode::serialize_into(checksum_slice, &checksum)?;
```
ikopylov, line 124:
```suggestion
        checksum_slice.copy_from_slice(&checksum.to_le_bytes());
```

src/record.rs
```diff
@@ -80,6 +80,52 @@ impl Meta {
...
+            mut buf,
+            offset_pos,
+            header_checksum_pos,
+            header_start,
+            header_len,
+        } = self;
+        let offset_slice = &mut buf[offset_pos..];
+        bincode::serialize_into(offset_slice, &blob_offset)?;
+        let header_slice = &buf[header_start..(header_start + header_len)];
+        let checksum = CRC32C.checksum(header_slice);
```
ikopylov, line 122: Explicit data type:
```suggestion
        let checksum: u32 = CRC32C.checksum(header_slice);
```

src/record.rs
```diff
@@ -80,6 +80,52 @@ impl Meta {
...
+        Self {
+            buf,
+            offset_pos,
+            header_checksum_pos,
+            header_start,
+            header_len,
+        }
+    }
+
+    pub(crate) fn into_serialized_with_header_checksum(
```
ikopylov, line 108: It is super critical to have a test for this function. It should compare serialization results between old and new aproaches

src/record.rs
```diff
@@ -80,6 +80,52 @@ impl Meta {
...
+        self,
```
ikopylov, line 109: Looks like `mut self` will help to avoid copy at the beginning of the function

src/record.rs
idruzhitskiy: Done

src/record.rs
idruzhitskiy: There is a test for header now, which is the only new logic

src/record.rs
idruzhitskiy: It panics at `source slice length does not match destination slice length`, which is not checked in serialization function (it just writes as much as possible). Do you think we should take len and obtain exact slice?

CHANGELOG.md
```diff
@@ -55,6 +55,9 @@ Pearl changelog
 - Change file operations to work on full buffers (#190)
 - Remove iterators from bloom filter (#194)
 - Excessive key conversion removed (#193)
+- Removed serialization from critical section (#188)
+- Removed multiple header checksum calculations (#206)
+- Rename functions to show that it returns last entry (#199)
```
ikopylov, line 60: Move these lines to `[Unreleased]` section

src/blob/core.rs
```diff
@@ -243,16 +244,23 @@ where
...
         debug!("blob write");
         // Only one upgradable_read lock is allowed at a time
+        let (partially_serialized, header) = record.to_partially_serialized_and_header()?;
         let blob = blob.upgradable_read().await;
-        Self::write_locked(blob, key, record).await
+        Self::write_locked(blob, key, partially_serialized, header).await
     }
 
-    async fn write_mut(&mut self, key: &K, mut record: Record) -> Result<RecordHeader> {
+    async fn write_mut(
```
ikopylov, line 252: I think there is no need to update `write_mut` signature. It is better to receive `Record` as it was before and only inside perform the conversion to `PartiallySerializedRecord`. 
The optimization with `PartiallySerializedRecord` was designed for `write` + `write_locked` path

src/record/mod.rs
```diff
@@ -1,10 +1,14 @@
+pub(crate) mod partially_serialized;
+
```
ikopylov, line 2: I think it is better to place all content to `record.rs` module and in `mod.rs` only re-export all required types

src/record/mod.rs
```diff
@@ -309,6 +281,19 @@ impl Header {
         bincode::serialized_size(&self).expect("calc record serialized size")
     }
 
+    pub(crate) fn set_offset_checksum(&mut self, blob_offset: u64, header_checksum: u32) {
+        self.blob_offset = blob_offset;
+        self.header_checksum = header_checksum;
+    }
+
+    fn blob_offset_offset(len: usize) -> usize {
```
ikopylov, line 289:
```suggestion
    const fn blob_offset_offset(len: usize) -> usize {
```

src/record/mod.rs
```diff
@@ -309,6 +281,19 @@ impl Header {
...
+        len - 24
+    }
+
+    fn checksum_offset(len: usize) -> usize {
```
ikopylov, line 293:
```suggestion
    const fn checksum_offset(len: usize) -> usize {
```

src/record/mod.rs
```diff
@@ -309,6 +281,19 @@ impl Header {
         bincode::serialized_size(&self).expect("calc record serialized size")
     }
 
+    pub(crate) fn set_offset_checksum(&mut self, blob_offset: u64, header_checksum: u32) {
+        self.blob_offset = blob_offset;
+        self.header_checksum = header_checksum;
```
ikopylov, line 286: Please, add `debug_assert` that verifies, that checksum is correct and equal to `update_checksum` result

src/record/mod.rs
```diff
@@ -309,6 +281,19 @@ impl Header {
...
+
+    fn blob_offset_offset(len: usize) -> usize {
+        len - 24
+    }
+
+    fn checksum_offset(len: usize) -> usize {
+        len - 4
+    }
+
     fn update_checksum(&mut self) -> bincode::Result<()> {
```
ikopylov, line 297: Return checksum value here

src/record/mod.rs
```diff
@@ -376,3 +361,32 @@ impl Header {
...
+#[cfg(test)]
+mod tests {
+    use bytes::Bytes;
+
+    use crate::prelude::*;
+
+    #[test]
+    pub fn partial_serialization_preserves_checksum_and_offset() -> Result<()> {
+        let key = vec![0, 0, 0];
+        let data = Bytes::new();
```
ikopylov, line 374: Please, don't use empty data. Init it with some values

src/record/mod.rs
```diff
@@ -376,3 +361,32 @@ impl Header {
...
+
+#[cfg(test)]
+mod tests {
+    use bytes::Bytes;
+
+    use crate::prelude::*;
+
+    #[test]
+    pub fn partial_serialization_preserves_checksum_and_offset() -> Result<()> {
+        let key = vec![0, 0, 0];
```
ikopylov, line 373: Please, run the test logic multiple times for different key and offset values.
```
for i in 0..8 {
   let key = vec![0, 0, i];
   ...
   let offset: u64 = i * 101;
   ...
}

File: src/record/partially_serialized.rs [hunk 21]
```diff
@@ -0,0 +1,98 @@
...
+        let Self {
+            mut buf,
+            header_len,
+        } = self;
+        let offset_pos = RecordHeader::blob_offset_offset(header_len);
+        let checksum_pos = RecordHeader::checksum_offset(header_len);
+        let offset_slice = &mut buf[offset_pos..];
+        bincode::serialize_into(offset_slice, &bytes_offset)?;
+        let header_slice = &buf[..header_len];
+        let checksum: u32 = CRC32C.checksum(header_slice);
```
ikopylov, line 62: Checksum should be set explicitly to 0 before CRC32 calculation. Otherwise the result can be incorrect

src/record.rs
ikopylov: Yes, the size of the type can be calculated with `std::mem::size_of::<type>()`:
```
offset_slice[..std::mem::size_of::<u64>()].copy_from_slice(&blob_offset.to_le_bytes());
```

src/record/partially_serialized.rs
```diff
@@ -0,0 +1,98 @@
...
+        let data = if head.len() + data.len() > MAX_SINGLE_PASS_DATA_SIZE {
+            PartiallySerializedData::DoublePass(head.freeze(), data)
+        } else {
+            head.extend_from_slice(&data);
+            PartiallySerializedData::SinglePass(head.freeze())
+        };
+        Ok((data, checksum))
+    }
+
+    pub async fn write_to_file(self, file: &File, offset: u64) -> Result<(u64, u32)> {
```
ikopylov, line 94: Please, declare a struct for result, that will have named fields. With tuple it is harder to understand what is the meaning of every value

src/record/partially_serialized.rs
```diff
@@ -0,0 +1,98 @@
...
+    fn into_serialized_with_header_checksum(
+        self,
+        blob_offset: u64,
+    ) -> Result<(PartiallySerializedData, u32)> {
+        let Self { head, data } = self;
+        let (mut head, checksum) = head.finalize_with_checksum(blob_offset)?;
+        let data = if head.len() + data.len() > MAX_SINGLE_PASS_DATA_SIZE {
+            PartiallySerializedData::DoublePass(head.freeze(), data)
+        } else {
+            head.extend_from_slice(&data);
```
ikopylov, line 88: Reallocation happens here. You lost the optimization that allows to allocate the buffer of the correct size only once

src/record.rs
ikopylov: I think an additional test required, that will compare raw bytes after serialization with old and with new approach

src/record/partially_serialized.rs
ikopylov: It seems that size based separation would need to be done at the Record level. And `PartiallySerializedRecord` in this case would look like:
```
struct PartiallySerializedRecord {
    header_containing_buf: BytesMut,
    header_len: usize,
    data_buf: Option<Bytes>
}
```
And all 3 structs would be combined into that `PartiallySerializedRecord`.

If you see the better way to avoid reallocation, then please use it

src/record.rs
idruzhitskiy: So we keep old approach just for the tests?

src/record/record.rs
```diff
@@ -309,10 +288,25 @@ impl Header {
...
 
-    fn update_checksum(&mut self) -> bincode::Result<()> {
+    pub(crate) fn set_offset_checksum(&mut self, blob_offset: u64, header_checksum: u32) {
+        self.blob_offset = blob_offset;
+        self.header_checksum = 0;
+        debug_assert_eq!(self.crc32().expect("debug assert crc32"), header_checksum);
+        self.header_checksum = header_checksum;
+    }
+
+    pub(super) const fn blob_offset_offset(len: usize) -> usize {
```
ikopylov, line 298: Please, add comment, that this function calculates offset to the `offset` field in serialized header. And that `len` is the length of serialized header in bytes

src/record/record.rs
```diff
@@ -309,10 +288,25 @@ impl Header {
...
+        len - 24
+    }
+
+    pub(super) const fn checksum_offset(len: usize) -> usize {
```
ikopylov, line 302: Add comment here too

src/record/partially_serialized.rs
```diff
@@ -0,0 +1,111 @@
...
+    fn finalize_with_checksum(
+        mut buf: BytesMut,
+        header_len: usize,
+        bytes_offset: u64,
+    ) -> Result<(BytesMut, u32)> {
+        use std::mem::size_of;
+
+        let offset_pos = RecordHeader::blob_offset_offset(header_len);
+        let checksum_pos = RecordHeader::checksum_offset(header_len);
+        let offset_slice = &mut buf[offset_pos..];
```
ikopylov, line 77: I think, `offset_slice` can be limited to the correct length right here.
```
let offset_slice = &mut buf[offset_pos..(offset_pos + size_of::<u64>())];
```

src/record/partially_serialized.rs
```diff
@@ -0,0 +1,111 @@
...
+        offset_slice[..size_of::<u64>()].copy_from_slice(&bytes_offset.to_le_bytes());
+
+        let checksum_slice = &mut buf[checksum_pos..];
```
ikopylov, line 80: Same here

src/record/partially_serialized.rs
```diff
@@ -0,0 +1,111 @@
...
+        checksum_slice[..size_of::<u32>()].copy_from_slice(&0u32.to_le_bytes());
+
+        let header_slice = &buf[..header_len];
+        let checksum: u32 = CRC32C.checksum(header_slice);
+        let checksum_slice = &mut buf[checksum_pos..];
+        checksum_slice[..size_of::<u32>()].copy_from_slice(&checksum.to_le_bytes());
```
ikopylov, line 86: You can reuse `checksum_slice` created above

src/record/partially_serialized.rs
```diff
@@ -0,0 +1,111 @@
...
+        res.map(|l| PartiallySerializedWriteResult {
+            bytes_written: l,
+            header_checksum: checksum,
+        })
+    }
+
+    fn finalize_with_checksum(
+        mut buf: BytesMut,
+        header_len: usize,
+        bytes_offset: u64,
```
ikopylov, line 71: `blob_offset` or `record_offset` name would be better

src/record.rs
ikopylov: There is no need to keep the old approach. You can reproduce the old serialization logic right inside the test function. This will be a backward compatibility test. You need to create a record and make its clone. Serialize one copy using the old approach: set the offset, update the checksum, write the header, metadata and data to the buffer. Then serialize the second copy by calling new functions. Finally, compare the resulting arrays

src/record/partially_serialized.rs
idruzhitskiy: No, since it's mutable borrow can't cross immutable borrow in header slice

src/record/record.rs
```diff
@@ -376,3 +375,71 @@ impl Header {
...
+            let mut new_method_buf = BytesMut::with_capacity(header_size + data.len());
+            new_method_buf.extend(&head);
+            if let Some(body) = body {
+                new_method_buf.extend(&body);
+            }
+
+            let mut old_method_buf = BytesMut::with_capacity(header_size + data.len());
+            header.blob_offset = offset;
+            header.header_checksum = 0;
+            header.header_checksum = header.crc32()?;
```
ikopylov, line 436: Use `header.update_checksum` instead

ikopylov (approved):
LGTM

PR #223, opened 2022-12-04
215 validate checksum in index regen

### LINKED ISSUE #215: Validate record check sum during index regeneration
Currently, there is no validation. That hides corruption errors.

Boneyan:
Index regeneration checks record data checksum
Closes #215

### DIFF
--- a/src/blob/core.rs
+++ b/src/blob/core.rs
@@ -198,12 +198,12 @@ where
         Ok(blob)
     }
 
-    async fn raw_records(&self) -> Result<RawRecords> {
+    async fn raw_records(&self, validate_data_during_index_regen: bool) -> Result<RawRecords> {
         RawRecords::start(
             self.file.clone(),
             bincode::serialized_size(&self.header)?,
             K::LEN as usize,
-            self.validate_data_during_index_regen,
+            validate_data_during_index_regen,
         )
         .await
         .context("failed to create iterator for raw records")
@@ -216,7 +216,8 @@ where
             return Ok(());
         }
         debug!("index file missed");
-        let raw_r = self.raw_records().await.with_context(|| {
+        let raw_r = self.raw_records(self.validate_data_during_index_regen)
+        .await.with_context(|| {
             format!(
                 "failed to read raw records from blob {:?}",
                 self.name.to_path()
--- a/src/blob/core.rs
+++ b/src/blob/core.rs
@@ -29,7 +29,7 @@ where
     file: File,
     current_offset: u64,
     created_at: SystemTime,
-    validate_data_in_regen: bool,
+    validate_data_during_index_regen: bool,
 }
 
 impl<K> Blob<K>
@@ -58,7 +58,7 @@ where
             file,
             current_offset: 0,
             created_at: SystemTime::now(),
-            validate_data_in_regen: config.validate_data_in_regen(),
+            validate_data_during_index_regen: config.validate_data_during_index_regen(),
         };
         blob.write_header().await?;
         Ok(blob)
@@ -178,7 +178,7 @@ where
             index,
             current_offset: size,
             created_at,
-            validate_data_in_regen: config.validate_data_in_regen(),
+            validate_data_during_index_regen: config.validate_data_during_index_regen(),
         };
         trace!("call update index");
         if is_index_corrupted || size as u64 > header_size {
@@ -203,7 +203,7 @@ where
             self.file.clone(),
             bincode::serialized_size(&self.header)?,
             K::LEN as usize,
-            self.validate_data_in_regen,
+            self.validate_data_during_index_regen,
         )
         .await
         .context("failed to create iterator for raw records")
@@ -498,11 +498,11 @@ struct RawRecords {
     current_offset: u64,
     record_header_size: u64,
     file: File,
-    validate_data: bool,
+    validate_data_during_index_regen: bool,
 }
 
 impl RawRecords {
-    async fn start(file: File, blob_header_size: u64, key_size: usize, validate_data: bool) -> Result<Self> {
+    async fn start(file: File, blob_header_size: u64, key_size: usize, validate_data_during_index_regen: bool) -> Result<Self> {
         let current_offset = blob_header_size;
         debug!("blob raw records start, current offset: {}", current_offset);
         let size_of_len = bincode::serialized_size(&(0_usize))? as usize;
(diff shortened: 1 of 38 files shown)


src/blob/core.rs
```diff
@@ -538,8 +538,11 @@ impl RawRecords {
         debug!("blob raw records load");
         let mut headers = Vec::new();
         while self.current_offset < self.file.size() {
-            let header = self.read_current_record_header().await.with_context(|| {
-                format!("read record header failed, at {}", self.current_offset)
+            let (header, data) = self.read_current_record_header_and_data().await.with_context(|| {
```
ikopylov, line 541: This should be configurable behavior: configuration parameter value should come from `node.yaml`. By default checksum validation should be disabled, but can be enabled through config

src/blob/index/core.rs
```diff
@@ -26,13 +28,15 @@ impl IndexParams {
 pub struct IndexConfig {
     pub bloom_config: Option<BloomConfig>,
     pub recreate_index_file: bool,
+    pub validate_data_in_regen: bool,
```
ikopylov, line 31: `IndexConfig` is not suitable for this parameter. This parameter is for data validation (not index validation). So you should add it to storage config (https://github.com/qoollo/pearl/blob/master/src/storage/config.rs)

src/blob/core.rs
```diff
@@ -541,13 +542,18 @@ impl RawRecords {
         }
     }
 
-    async fn load(mut self) -> Result<Option<Vec<RecordHeader>>> {
+    async fn load(mut self, validate_data: bool) -> Result<Option<Vec<RecordHeader>>> {
```
ikopylov, line 545: From design point of view, it is better to pass this argument to `start` function and store its value as a field in `RawRecords` struct

src/blob/core.rs
```diff
@@ -557,7 +563,7 @@ impl RawRecords {
         }
     }
 
-    async fn read_current_record_header(&mut self) -> Result<RecordHeader> {
+    async fn read_current_record(&mut self, read_data: bool) -> Result<(RecordHeader, Vec<u8>)> {
```
ikopylov, line 566: It is better to return `Result<(RecordHeader, Option<Vec<u8>>)>` to explicitly indicate that data was not read

src/blob/index/core.rs
Boneyan: So, it is also shouldn't be part of IndexParams? It is ok to store this variable as Blob struct field?

ikopylov: Yes, it is ok

src/blob/core.rs
```diff
@@ -56,6 +58,7 @@ where
             file,
             current_offset: 0,
             created_at: SystemTime::now(),
+            validate_data_in_regen: config.validate_data_in_regen(),
```
ikopylov, line 61: `validate_data_during_index_regen`

ikopylov: Please, change name everywhere

src/blob/core.rs
```diff
@@ -573,8 +586,19 @@ impl RawRecords {
             })?;
         self.current_offset += self.record_header_size;
         self.current_offset += header.meta_size();
+        let data =
+        if read_data {
```
ikopylov, line 590: Please add indentation so it can be seen that this is part of the value assignment to `data`

src/blob/core.rs
```diff
@@ -573,8 +586,19 @@ impl RawRecords {
...
+            buf.resize(header.data_size() as usize, 0);
+            self.file
+                .read_at(&mut buf, self.current_offset)
```
ikopylov, line 593: `read_exact_at`

src/blob/core.rs
```diff
@@ -557,7 +570,7 @@ impl RawRecords {
         }
     }
 
-    async fn read_current_record_header(&mut self) -> Result<RecordHeader> {
+    async fn read_current_record(&mut self, read_data: bool) -> Result<(RecordHeader, Option<Vec<u8>>)> {
         let mut buf = vec![0; self.record_header_size as usize];
         self.file
             .read_at(&mut buf, self.current_offset)
```
ikopylov, line 576: `read_exact_at`

src/blob/core.rs
Boneyan: file is not std File struct, but File struct from pearl, as I understand read_at() already will read exact buf.len() bytes

ikopylov: Ok. Then it is alright

src/blob/core.rs
```diff
@@ -198,6 +203,7 @@ where
             self.file.clone(),
             bincode::serialized_size(&self.header)?,
             K::LEN as usize,
+            self.validate_data_during_index_regen,
```
ikopylov, line 206: I suggest to pass this as a parameter to `raw_records`. If someone will start using this function just to read record headers, it would be surprise, that it also validates data. So it is better to give explicit control over the behavior though parameter

src/blob/core.rs
```diff
@@ -198,12 +198,12 @@ where
         Ok(blob)
     }
 
-    async fn raw_records(&self) -> Result<RawRecords> {
+    async fn raw_records(&self, validate_data_during_index_regen: bool) -> Result<RawRecords> {
```
ikopylov, line 201: Here it is better to name the argument `validate_data`, because `raw_records` is not directly related to regenerating indexes

src/blob/core.rs
```diff
@@ -492,10 +499,11 @@ struct RawRecords {
     current_offset: u64,
     record_header_size: u64,
     file: File,
+    validate_data_during_index_regen: bool,
 }
 
 impl RawRecords {
-    async fn start(file: File, blob_header_size: u64, key_size: usize) -> Result<Self> {
+    async fn start(file: File, blob_header_size: u64, key_size: usize, validate_data_during_index_regen: bool) -> Result<Self> {
```
ikopylov, line 506: Here the same: just `validate_data`

src/blob/core.rs
```diff
@@ -492,10 +499,11 @@ struct RawRecords {
     current_offset: u64,
     record_header_size: u64,
     file: File,
+    validate_data_during_index_regen: bool,
```
ikopylov, line 502: And here

ikopylov (approved):
LGTM

PR #226, opened 2022-12-24
225 update crc

### LINKED ISSUE #225: Update `crc` crate version

lostl1ght:
Closes #225

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ bincode = "1.3"
 bitvec = "1.0"
 chrono = "0.4"
 clap = { version = "3.2", optional = true }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -30,7 +30,7 @@ bincode = "1.3"
 bitvec = "1.0"
 chrono = "0.4"
 clap = { version = "3.2", optional = true }
-crc = "=2.0.0"
+crc = "3.0"
```
ikopylov, line 33: It is better to require concrete version of crate, because it theoretically can change calculation logic. Every update should be validated explicitly
```suggestion
crc = "=3.0.0"
```

PR #232, opened 2023-01-14
231 pass and store timestamp within record

### LINKED ISSUE #229: Make it able to pass metadata to the deletion record

### LINKED ISSUE #231: Pass and store user timestamp within record, return it instead of creation timestamp
Related issue: https://github.com/qoollo/bob/issues/708

### LINKED ISSUE #241: Decide the best way of treating errors in `Storage::contains` and `Storage::read`
Right now, `read` skips errors from BLOB and continue reading from the next one. `contains` otherwise propagates the first error.
Looks like it is better to stop on the first error to preserve correct versioning of data. But this should be investigated

### LINKED ISSUE #276: `Storage::read_with_optional_meta` can load Entries first, select the latest one and only then load the data
Right now `Storage::read_with_optional_meta` loads data for all records for specified key. This can be improved by loading entries first, and then loading data for only one `Entry`

ikopylov:
Closes #231 
Closes #229 
Closes #276 
Closes #241


src/blob/core.rs
```diff
@@ -262,28 +262,22 @@ where
         Ok(header)
     }
 
-    pub(crate) async fn read_last(
+    #[allow(dead_code)]
```
idruzhitskiy, line 265: Why this allow is needed?

src/blob/entry.rs
```diff
@@ -31,12 +32,15 @@ impl Entry {
...
-            .context("Record load failed")?;
+            .with_context(|| format!("Record load failed from BLOB: {}", self.blob_file_name.to_path().display()))?;
         let mut buf = buf.freeze();
         let data_buf = buf.split_off(meta_size);
-        let meta = Meta::from_raw(&buf).map_err(|err| Error::from(err))?;
-        let record = Record::new(self.header.clone(), meta, data_buf);
-        record.validate()
+        let meta = Meta::from_raw(&buf)
+            .map_err(|err| Error::from(err))
+            .with_context(|| format!("Deserialization failed for Meta loaded from BLOB: {}", self.blob_file_name.to_path().display()))?;
```
idruzhitskiy, line 40: I think most of the error messages starts with lowercase in pearl

src/storage/core.rs
```diff
@@ -997,12 +992,23 @@ where
...
     /// Fails after any disk IO errors.
-    pub async fn delete(&self, key: impl AsRef<K>, only_if_presented: bool) -> Result<u64> {
+    pub async fn delete(&self, key: impl AsRef<K>, timestamp: BlobRecordTimestamp, only_if_presented: bool) -> Result<u64> {
+        self.delete_with_optional_meta(key, timestamp, None, only_if_presented).await
+    }
+
+    /// Delete entries with matching key. Appends metadata to deletion record
+    /// # Errors
+    /// Fails after any disk IO errors.
+    pub async fn delete_with(&self, key: impl AsRef<K>, timestamp: BlobRecordTimestamp, meta: Meta, only_if_presented: bool) -> Result<u64> {
```
idruzhitskiy, line 1002: Isn't this method redundant, it just accepts subset of `delete_with_optional_meta` arguments

src/blob/index/core.rs
```diff
@@ -332,7 +332,17 @@ where
                 let records_allocated;
                 if let Some(v) = data.headers.get_mut(key) {
                     let old_capacity = v.capacity();
-                    v.push(h);
+                    // Keep ordered by timestamp
+                    let mut pos = 0;
+                    if v.len() > 4 {
```
idruzhitskiy, line 337: Does it provide any noticeable effects in benchmarks on pearl?

src/blob/core.rs
ikopylov: `read_latest` is no longer used. Since this request `Storage` uses `get_latest_entry`. I thought it might be worth keeping this function for the future. What do you think?

src/blob/entry.rs
ikopylov: I don't think it's critical. However, I've changed this

src/storage/core.rs
ikopylov: I keep it consistent with the `write` functions. We also have 3 of them: public `write`, public `write_with` and internal `write_with_optional_meta`

src/blob/index/core.rs
ikopylov: In pearl benchmarks, this change is not noticable. However, from my experience: in micro-benchmarks linear search in arrays of small length is faster than binary search due to better cache usage

src/blob/core.rs
idruzhitskiy: I think it's better to just keep in git history since it can make it harder to refactor code

idruzhitskiy (approved):
LGTM

src/blob/core.rs
ikopylov: I've removed this function

idruzhitskiy: What about other dead_code allowances, they should be there?

ikopylov: If we assume that the only way to read data now is by getting `Entry` first, then we should keep `read_all_entries`. 
`contains` is questionable, but as long as it is implemented as a simple redirection to `get_latest_entry`, we can keep it. It is good to have such function just as a declaration, that the struct supports fast checking for existence.

ikopylov: I removed all dead code

PR #240, opened 2023-01-17
233 unexpected eof conversion

### LINKED ISSUE #233: Handle `failed to fill whole buffer` in blob corruption detection logic
Starting happened after changing read methods to `read_exact`:
```
2023-01-15 20:42:25  bob_common::configs::node:428 ERROR    can't initialize holder, attempt 3/3, error backend storage error: pearl error: failed to init from existing blobs

Caused by:
    0: failed to read blobs
    1: Failed to read existing blob: /opt/disk1/bob_perf/bob/0/1673776800_9JVKD4lftV/bob_9JVKD4lftV.424.blob
    2: failed to read blob header. Blob file: "/opt/disk1/bob_perf/bob/0/1673776800_9JVKD4lftV/bob_9JVKD4lftV.424.blob"
    3: failed to read from file: bob_9JVKD4lftV.424.blob
    4: failed to fill whole buffer
```

Specific BLOB file had 0 size

Boneyan:
Converting UnexpectedEOF errors to bincode errors before deserialize() calls
Closes #233


src/error.rs
```diff
@@ -174,3 +179,23 @@ impl AsPearlError for anyhow::Error {
...
+    if let Some(io_error) = e.downcast_ref::<std::io::Error>() {
+        if let std::io::ErrorKind::UnexpectedEof = io_error.kind() {
+            return AHError::from(Error::bincode("unexpected eof".into()));
+        }
+    }
+    e
+}
+
+/// Util function for UnexpectedEOF error conversion with additional context
+pub fn unexpected_eof_converter_ctx(e: AHError, context: String) -> AHError {
```
ikopylov, line 194: I think, this function is not needed. It is better to use `with_context` to set the context after the conversion of error happened

src/lib.rs
```diff
@@ -64,7 +64,7 @@ pub use filter::{Bloom, BloomDataProvider, BloomProvider, Config as BloomConfig,
 pub mod tools;
 
 pub use blob::Entry;
-pub use error::{Error, Kind as ErrorKind};
+pub use error::{Error, Kind as ErrorKind, unexpected_eof_converter, unexpected_eof_converter_ctx};
```
ikopylov, line 67: `unexpected_eof_converter` and `unexpected_eof_converter_ctx` should not be exposed outside of the crate!

tests/tests.rs
```diff
@@ -156,6 +157,122 @@ async fn test_multithread_read_write() -> Result<(), String> {
...
+                            map.insert(key, usize::MAX);
+                        },
+                        1 | 2 | 3 => {
+                            let data_len = rng.gen::<usize>() % data.len();
+                            write_one(&st, key, &data[0..data_len], None).await.expect("write success");
+                            map.insert(key, data_len);
+                        },
+                        4 | 5 | 6 => {
+                            let exist_res = st.contains(KeyTest::new(key)).await.expect("exist success");
+                            match exist_res {
```
idruzhitskiy, line 201: I think this check can be extracted into separate function (with flag for checking size in `found`), so test will be more concise

src/error.rs
```diff
@@ -175,3 +187,35 @@ impl AsPearlError for anyhow::Error {
...
+pub(crate) trait IntoBincodeIfUnexpectedEofTrait {
+    /// Converts [`IOErrorKind::UnexpectedEof`] into [`Kind::Bincode`] error wrapped into [`anyhow::Error`].
+    /// Should be called when further deserialization expected
+    fn into_bincode_if_unexpected_eof(self) -> anyhow::Error;
+}
+
+impl IntoBincodeIfUnexpectedEofTrait for IOError {
+    fn into_bincode_if_unexpected_eof(self) -> anyhow::Error {
+        if self.kind() == IOErrorKind::UnexpectedEof {
+            Error::bincode("Can't read whole buffer, required for deserialization due to unexpected end of file".to_string()).into()
```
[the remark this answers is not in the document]
ikopylov: It seems that in such a case it is better to declare a constant, rather than create a function

tests/tests.rs
ikopylov: It is not that easy, due to different result types. `contains` returns `ReadResult<BlobRecordTimestamp>`, `read` returns `ReadResult<Bytes>`. 

Other than that, `assert` inside will also print the line number, by which it is easier to diagnose the cause. If all checks are wrapped in a function, then the line number will not tell you anything

ikopylov: I've changed the test code to compare `ReadResult<>` in `assert_eq!`. This required adding several helper function: one to map `Option<>` into `ReadResult<>`, other to reset timestamp inside `ReadResult::Deleted()`

idruzhitskiy (approved):
LGTM

PR #242, opened 2023-01-23
234 background fsync

### LINKED ISSUE #234: Significant `put` performance drop due to `fsync_all` call on blob close
We must find the best way to sync data. Perhaps we can run `fsync_all` in background. Or run `fsync_all` periodically.
Both works.

Boneyan:
Blob performes fsync if buffered bytes are larger than max_dirty_bytes_before_sync config param
Closes #234

### DIFF
--- a/src/blob/file.rs
+++ b/src/blob/file.rs
@@ -18,6 +18,7 @@ pub struct File {
     ioring: Option<Rio>,
     no_lock_fd: Arc<StdFile>, // requires only for read_at/write_at methods
     size: Arc<AtomicU64>,
+    dirty_bytes: Arc<AtomicUsize>,
 }
 
 #[derive(PartialEq, Eq)]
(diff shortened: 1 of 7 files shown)


ikopylov:
Please, do the following task here too: https://github.com/qoollo/pearl/issues/243

src/blob/file.rs
```diff
@@ -255,18 +268,34 @@ impl File {
...
+        };
+        self.dirty_bytes.store(0, ORD);
+        Ok(())
+    }
+
+    pub(crate) fn background_fsyncdata(&self) {
+        let fd = self.no_lock_fd.clone();
+        if let Some(ref ioring) = self.ioring {
+            let ioring = ioring.clone();
+            tokio::spawn(async move {
```
ikopylov, line 292: We should track the spawning task somewhere. The best place for this is `ObserverWorker`

ikopylov:
This PR should be rewritten. `fsync` task should be started from `ObserverWorker` and `ObserverWorker` should track its completion

ikopylov:
Replaced by https://github.com/qoollo/pearl/pull/288

PR #244, opened 2023-02-08
241 skip errors on contains

### LINKED ISSUE #241: Decide the best way of treating errors in `Storage::contains` and `Storage::read`

ikopylov:
Closes #241

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ Pearl changelog
 
 
 #### Changed
-- Skip errors in Storage::contains (#241)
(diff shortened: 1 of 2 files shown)


ikopylov:
Probably, it is better not to skip the errors. So instead of changing `contains` logic, it is better to update `get` logic

ikopylov:
Strategy should be re-decided

PR #250, opened 2023-03-07
228 cancellation protection

### LINKED ISSUE #228: Partially fix data corruption due to canceling an asynchronous operation
Unfortunately, async operation cancellation can corrupt data on disk and break the consistency of data in memory. We can't fix this completely, because cancellation can happen at any point where `.await` is used. But we can improve the situation and prevent data corruption on disk:
1. We track offset within file in 2 places: in Blob (https://github.com/qoollo/pearl/blob/master/src/blob/core.rs#L30) and in File (https://github.com/qoollo/pearl/blob/master/src/blob/file.rs#L14). They are easily can diverge. We can choose not to track `offset` on Blob level and only track it on File level
2. Update File offset at the place where cancellation is not possible. Currently it updates here: https://github.com/qoollo/pearl/blob/master/src/blob/file.rs#L62 and right on the next line we use `.await`. We have to move the offset update right into place before the synchronous write happens: https://github.com/qoollo/pearl/blob/master/src/blob/file.rs#L96
3. We need to set an offset in the record header (https://github.com/qoollo/pearl/blob/master/src/blob/core.rs#L274), but with new approach it becomes harder. We need to transfer a special structure directly to the place where the data is written to the file, which allows to patch the data with the `offset` value. Such struct will be added in this PR: https://github.com/qoollo/pearl/pull/220. Also we should return the offset value to the caller, so it can also change value within header struct

idruzhitskiy:
Closes #228


src/io/unix/sync.rs
```diff
@@ -59,39 +59,50 @@ impl File {
...
-        first_buf: Bytes,
-        second_buf: Bytes,
-    ) -> IOResult<()> {
-        let first_len = first_buf.len() as u64;
-        let second_len = second_buf.len() as u64;
-        let len = first_len + second_len;
+        c: impl WritableDataCreator<R>,
+        len: u64,
+    ) -> Result<R> {
         let mut offset = self.size.fetch_add(len, Ordering::SeqCst);
```
ikopylov, line 67: The `size` update must be inside the block of code passed to `inplace_sync_call` or `background_sync_call`. This is because there is an `await` below that can be the point of cancellation

src/io/unix/sync.rs
```diff
@@ -59,39 +59,50 @@ impl File {
...
+                };
+                Ok(data)
             })
             .await
         }
     }
 
+    pub(crate) async fn write_append_all(&self, buf: Bytes) -> IOResult<()> {
+        let offset = self.size.fetch_add(buf.len() as u64, Ordering::SeqCst);
+        self.write_all_at(offset, buf).await
```
ikopylov, line 103: I think it would  be good to update this function as well: perform `self.size.fetch_add` within the code block passed to `inplace_sync_call` or to `background_sync_call`. This will require to manually inline the code from `write_all_at` to this function, but I think this is not a big problem

src/io/unix/sync.rs
```diff
@@ -59,39 +59,50 @@ impl File {
...
+    pub(crate) async fn write_append_writable_data<R: Send + 'static>(
         &self,
-        first_buf: Bytes,
-        second_buf: Bytes,
-    ) -> IOResult<()> {
-        let first_len = first_buf.len() as u64;
-        let second_len = second_buf.len() as u64;
-        let len = first_len + second_len;
+        c: impl WritableDataCreator<R>,
+        len: u64,
```
ikopylov, line 65: `len` should not be passed as argument, it should be calculated within this function

src/io/unix/iouring.rs
```diff
@@ -69,31 +69,35 @@ impl File {
         self.sync.size()
     }
 
-    pub(crate) async fn write_append_all(&self, buf: Bytes) -> IOResult<()> {
+    pub(crate) async fn write_append_writable_data<R: Send + 'static>(
+        &self,
+        c: impl WritableDataCreator<R>,
+        len: u64,
```
ikopylov, line 75: Do not pass `len` as argument. This is not safe. You should calculate `len` within this function

src/io/mod.rs
```diff
@@ -7,3 +10,18 @@ pub use unix::IoDriver;
 
 #[cfg(not(target_family = "unix"))]
 compile_error!("Specified target platform is not supported (only unix family supported)");
+
+pub(crate) enum WritableData {
```
ikopylov, line 14: You can add `len()` function to this `enum`

ikopylov, line 14: You can add `len()` function to this enum

src/io/mod.rs
```diff
@@ -7,3 +10,18 @@ pub use unix::IoDriver;
...
+    Single(Bytes),
+    Double(Bytes, Bytes),
+}
+
+pub(crate) trait WritableDataCreator<R>:
+    FnOnce(u64) -> Result<(WritableData, R)> + Send + 'static
```
ikopylov, line 20: I don't like the approach with delegate. It is better to declare the function within this trait and call it

src/record/partially_serialized.rs
```diff
@@ -87,25 +103,4 @@ impl PartiallySerializedRecord {
...
-        let len = buf.len() as u64;
-        Self::process_file_result(file.write_append_all(buf).await).map(|_| len)
-    }
-
-    async fn write_double_pass(head: Bytes, data: Bytes, file: &File) -> Result<u64> {
-        let len = (head.len() + data.len()) as u64;
-        Self::process_file_result(file.write_append_all_buffers(head, data).await).map(|_| len)
-    }
-
-    fn process_file_result(result: std::result::Result<(), std::io::Error>) -> Result<()> {
```
ikopylov, line 101: You lost this result conversion! It's important to keep it because there is logic up the call chain that depends on this processing

src/io/mod.rs
idruzhitskiy: I can't: offset is needed to produce this data, and we can't obtain offset without len. So len will be part of the creator trait

ikopylov: I meant, that you can calculate it by the data within this enum:
```
impl WritableData {
    pub(crate) fn len(&self) -> usize {
         match self {
            // ....
         }
    }
}
```

src/io/unix/sync.rs
```diff
@@ -59,39 +59,61 @@ impl File {
...
+            _ => e.into(),
+        })
+    }
+
+    pub(crate) async fn write_append_all(&self, buf: Bytes) -> IOResult<()> {
+        let file = self.no_lock_fd.clone();
+        let size = self.size.clone();
+        if Self::can_run_inplace(buf.len() as u64) {
+            let offset = size.fetch_add(buf.len() as u64, Ordering::SeqCst);
+            debug_assert!(offset + buf.len() as u64 <= self.size());
```
ikopylov, line 108: This check is not needed here because `size` is updated one line before by exactly the size of the buffer

src/io/unix/sync.rs
```diff
@@ -59,39 +59,61 @@ impl File {
...
+            Self::inplace_sync_call(move || file.write_all_at(&buf, offset))
+        } else {
+            let offset = size.fetch_add(buf.len() as u64, Ordering::SeqCst);
+            debug_assert!(offset + buf.len() as u64 <= self.size());
```
ikopylov, line 112: Same here

src/io/unix/sync.rs
```diff
@@ -59,39 +59,61 @@ impl File {
...
+            }
+            _ => e.into(),
+        })
+    }
+
+    pub(crate) async fn write_append_all(&self, buf: Bytes) -> IOResult<()> {
+        let file = self.no_lock_fd.clone();
+        let size = self.size.clone();
+        if Self::can_run_inplace(buf.len() as u64) {
+            let offset = size.fetch_add(buf.len() as u64, Ordering::SeqCst);
```
ikopylov, line 107: I asked to move this update inside code block passed to `inplace_sync_call`

src/io/unix/sync.rs
```diff
@@ -59,39 +59,61 @@ impl File {
...
+            debug_assert!(offset + buf.len() as u64 <= self.size());
+            Self::inplace_sync_call(move || file.write_all_at(&buf, offset))
+        } else {
+            let offset = size.fetch_add(buf.len() as u64, Ordering::SeqCst);
```
ikopylov, line 111: Same here: this update should be done inside code block passed to `background_sync_call`

src/io/unix/iouring.rs
```diff
@@ -69,31 +69,35 @@ impl File {
...
+    pub(crate) async fn write_append_writable_data<R: Send + 'static>(
+        &self,
+        c: impl WritableDataCreator<R>,
+    ) -> Result<R> {
         if let Some(ref rio) = self.rio {
-            let offset = self.sync.file_size_append(buf.len() as u64);
-            self.write_all_at_aio(offset, buf, rio).await
+            let len = c.len();
+            let mut offset = self.sync.file_size_append(len);
+            let (res, data) = c(offset)?;
```
ikopylov, line 79: Compilation error here

src/record/partially_serialized.rs
```diff
@@ -3,20 +3,56 @@ use crate::prelude::*;
...
     }
 
     pub(crate) fn header_checksum(&self) -> u32 {
         self.header_checksum
     }
 }
 
+struct RecordWritableDataCreator {
+    record: PartiallySerializedRecord,
+    len: u64,
```
ikopylov, line 22: I think, it is better to calculate `len` on the fly in the `len()` function, not to store it here. This is because the overhead of copying of this additional 8 bytes field can be larger than just calculating the value on the fly

src/record/partially_serialized.rs
```diff
@@ -3,20 +3,56 @@ use crate::prelude::*;
...
+}
+
+impl From<PartiallySerializedRecord> for RecordWritableDataCreator {
+    fn from(record: PartiallySerializedRecord) -> Self {
+        let len = record.head_with_data.len() as u64
+            + record.data.as_ref().map(|v| v.len()).unwrap_or(0) as u64;
```
ikopylov, line 28: Move this calculation into `len()` function

src/record/partially_serialized.rs
```diff
@@ -3,20 +3,56 @@ use crate::prelude::*;
...
+        Self { record, len }
+    }
+}
+
+impl WritableDataCreator<(u64, u32)> for RecordWritableDataCreator {
```
ikopylov, line 33: Looks like you can implement `WritableDataCreator<(u64, u32)>` directly on `PartiallySerializedRecord`

ikopylov: Additionally: you already have `PartiallySerializedWriteResult` that contains `offset` + `checksum`, so you can use that type instead of tuple

src/io/unix/sync.rs
```diff
@@ -59,39 +59,61 @@ impl File {
...
 
+    fn write_data(file: &StdFile, mut offset: u64, writable_data: WritableData) -> Result<()> {
+        let file_result: IOResult<()> = match writable_data {
+            WritableData::Single(bytes) => file.write_all_at(&bytes, offset),
+            WritableData::Double(b1, b2) => file.write_all_at(&b1, offset).and_then(|_| {
+                offset = offset + b1.len() as u64;
+                file.write_all_at(&b2, offset)
+            }),
+        };
+        file_result.map_err(|e| match e.kind() {
```
ikopylov, line 95: I think, it is better to return `IoResult<()>` from all the functions within `File` (like it was before) and this error processing should be on `PartiallySerializedRecord` level (again: like it was before)

src/io/unix/sync.rs
```diff
@@ -59,39 +59,61 @@ impl File {
...
+        let size = self.size.clone();
         let file = self.no_lock_fd.clone();
         if Self::can_run_inplace(len) {
             Self::inplace_sync_call(move || {
-                file.write_all_at(&first_buf, offset)?;
-                offset = offset + first_len;
-                file.write_all_at(&second_buf, offset)?;
-                Ok(())
+                let offset = size.fetch_add(len, Ordering::SeqCst);
+                let (res, data) = c.create(offset)?;
```
ikopylov, line 72: This call prevent us from returning `IoResult<>`, but as I see the current implementation never returns an error, so `create` signature can be updated to `fn create(self, offset: u64) -> (WritableData, R)`

src/record/partially_serialized.rs
```diff
@@ -3,20 +3,56 @@ use crate::prelude::*;
...
+    fn create(self, offset: u64) -> Result<(WritableData, (u64, u32))> {
+        let PartiallySerializedRecord {
+            head_with_data,
+            header_len,
+            data,
+        } = self.record;
+        let (head, checksum) =
+            PartiallySerializedRecord::finalize_with_checksum(head_with_data, header_len, offset)?;
```
ikopylov, line 41: `finalize_with_checksum` is never return `Err`, so its signature should be updated:
```
    fn finalize_with_checksum(
        mut buf: BytesMut,
        header_len: usize,
        blob_offset: u64,
    ) -> (BytesMut, u32)
```

src/io/mod.rs
```diff
@@ -1,3 +1,6 @@
+use anyhow::Result;
```
ikopylov, line 1: Warning: `unused import: 'anyhow::Result'`

src/record/partially_serialized.rs
```diff
@@ -85,27 +79,35 @@ impl PartiallySerializedRecord {
...
-        result.map_err(|e| -> anyhow::Error {
-            match e.kind() {
-                kind if kind == IOErrorKind::Other || kind == IOErrorKind::NotFound => {
-                    Error::file_unavailable(kind).into()
-                }
-                _ => e.into(),
-            }
-        })
+    fn len(&self) -> u64 {
+        self.head_with_data.len() as u64 + self.data.as_ref().map(|v| v.len()).unwrap_or(0) as u64
```
ikopylov, line 111:
```suggestion
        self.head_with_data.len() as u64 + self.data.as_ref().map_or(0, |v| v.len()) as u64
```

ikopylov (approved):
LGTM

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Pearl changelog
 
 
 #### Updated
+- Update offset only when write is imminent (#228)
```
ikopylov, line 16:
```suggestion
- Update offset only when write is imminent to prevent data corruption due to async Future cancellation (#228)
```

PR #252, opened 2023-03-13
230 optional iouring

### LINKED ISSUE #230: Make the `io_uring` feature optional so that it can be disabled in cargo.toml
This should allow running Bob not only on Linux

idruzhitskiy:
Closes #230


idruzhitskiy:
Closes #230

src/blob/index/benchmarks.rs
```diff
@@ -97,7 +92,7 @@ async fn benchmark_from_records() {
         println!("Test {}...", i + 1);
         let _ = FileIndexStruct::from_records(
             Path::new(FILEPATH),
-            ioring.clone(),
+            IODriver::default(),
```
ikopylov, line 95: This line is inside a loop, so it will lead to creation of multiple instances of `IODriver`. This is bad for the `Rio` case. I suggest to create single instance at the beginning of this function and pass its clones here

src/blob/index/benchmarks.rs
```diff
@@ -157,7 +147,7 @@ async fn benchmark_from_file() {
                 EXTENSION.to_owned(),
                 PathBuf::from(DIR),
             ),
-            ioring.clone(),
+            IODriver::default(),
```
ikopylov, line 150: And here

src/blob/index/benchmarks.rs
```diff
@@ -179,19 +169,14 @@ async fn benchmark_get_any() {
...
-    let ioring = ioring.ok();
     println!("Generating headers...");
     let headers = generate_headers(RECORDS_AMOUNT, KEY_MAPPER);
     let meta = generate_meta(META_SIZE);
 
     println!("Creating file index from headers...");
     let findex = FileIndexStruct::from_records(
         Path::new(FILEPATH),
-        ioring.clone(),
+        IODriver::default(),
```
ikopylov, line 179: It would be good to create `IODriver` at the beginning for the function where `Rio` created before

src/blob/index/benchmarks.rs
```diff
@@ -233,19 +218,14 @@ async fn benchmark_get_all() {
...
```
ikopylov, line 228: Same here

src/blob/index/bptree/tests.rs
```diff
@@ -154,7 +154,7 @@ async fn magic_byte_corruption() {
 
     let findex = BPTreeFileIndex::<KeyType>::from_file(
         FileName::from_path(&Path::new(filename)).expect("failed to create filename"),
-        None,
+        IODriver::default(),
```
ikopylov, line 157: `IODriver` creates multiple times in this function. I suggest to do it only once at the beginning

src/blob/mod.rs
```diff
@@ -1,14 +1,14 @@
...
+pub(crate) mod config;
 pub(crate) mod header;
 pub(crate) mod index;
-pub(crate) mod config;
 
 pub(crate) use self::core::BLOB_INDEX_FILE_EXTENSION;
 pub(crate) use self::core::{Blob, FileName};
 pub use self::entry::Entry;
-pub(crate) use self::file::File;
+//pub(crate) use self::file::File;
```
ikopylov, line 11: It is better to remove the commented line

src/error.rs
```diff
@@ -42,6 +42,7 @@ impl Error {
         Self::new(Kind::ActiveBlobExists)
     }
 
+    #[cfg(feature = "async-io-rio")]
```
ikopylov, line 45: I think, this is not good. If this generates warning, then it is better to mark it with `#[allow(dead_code)]`

src/storage/builder.rs
```diff
@@ -167,13 +167,6 @@ impl Builder {
         self
     }
 
-    /// Enables linux AIO with provided io_uring.
-    #[must_use]
-    pub fn enable_aio(mut self, ioring: Rio) -> Self {
```
ikopylov, line 172: There should be a function, that allows to set `IODriver`

src/storage/builder.rs
```diff
@@ -32,7 +32,7 @@ use super::prelude::*;
 #[derive(Default, Debug)]
 pub struct Builder {
     config: Config,
-    ioring: Option<Rio>,
+    iodriver: IODriver,
```
ikopylov, line 35: `IODriver::default()` creates `Rio` inside. This is bad, because `Rio` construction is not cheep. It is better to declare `iodriver` as `Option<IODriver>` and use `unwrap_or(|| IODriver::default())` when creating the Storage

src/tools/utils.rs
```diff
@@ -138,7 +138,9 @@ where
 {
     let headers = match header.version() {
         HEADER_VERSION => {
-            let index = BPTreeFileIndex::<K>::from_file(FileName::from_path(path)?, None).await?;
+            let index =
+                BPTreeFileIndex::<K>::from_file(FileName::from_path(path)?, IODriver::default())
```
ikopylov, line 142: Here the sync `IoDriver` should be created (if `async-io-rio` enabled, then `Rio` should be `None` inside)

src/tools/validation.rs
```diff
@@ -37,7 +36,9 @@ where
         .into());
     } else if header.version() == HEADER_VERSION {
         block_on(async {
-            let index = BPTreeFileIndex::<K>::from_file(FileName::from_path(path)?, None).await?;
+            let index =
+                BPTreeFileIndex::<K>::from_file(FileName::from_path(path)?, IODriver::default())
```
ikopylov, line 40: Same here. Sync `IoDriver` should be used

tests/common.rs
```diff
@@ -127,12 +127,6 @@ pub async fn create_test_storage(
         .set_filter_config(Default::default())
         .set_deferred_index_dump_times(MIN_DEFER_TIME, MAX_DEFER_TIME)
         .allow_duplicates();
-    let builder = if let Ok(ioring) = rio::new() {
-        builder.enable_aio(ioring)
```
ikopylov, line 131: You should explicitly pass `IoDriver` instance to the builder

tests/tests.rs
```diff
@@ -979,12 +979,6 @@ async fn test_blob_header_validation() {
         .max_data_in_blob(100_000)
         .set_filter_config(Default::default())
         .allow_duplicates();
-    let builder = if let Ok(ioring) = rio::new() {
-        builder.enable_aio(ioring)
```
ikopylov, line 983: Same here

src/io/unix/iouring.rs
```diff
@@ -0,0 +1,176 @@
+use super::sync::{File as SyncFile, IODriver as SyncDriver};
+use crate::prelude::*;
+use bytes::{Bytes, BytesMut};
+use rio::Rio;
+use std::time::SystemTime;
+
+/// IO driver for file operations with optional async support
+#[derive(Debug, Clone)]
+pub struct IODriver {
```
ikopylov, line 9: `IoDriver` is  better (second letter is in lowercase)

src/io/unix/iouring.rs
```diff
@@ -0,0 +1,176 @@
...
+    rio: Option<Arc<Rio>>,
```
ikopylov, line 10: According to the source code `Rio` is a wrapper around `Arc<>` (https://docs.rs/rio/latest/src/rio/io_uring/uring.rs.html#5), so no need to wrap it once more with `Arc<>`

src/io/unix/iouring.rs
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    pub(crate) async fn create(&self, path: impl AsRef<Path>) -> IOResult<File> {
+        let sync = self.sync_driver.create(path).await?;
+        File::from_file(sync, self.rio.clone()).await
+    }
+}
+
+impl Default for IODriver {
+    fn default() -> Self {
```
ikopylov, line 35: I think, it is a bad idea to have `default` for `IoDriver`. It is better to have multiple explicit constructors to make all intentions clear:
1. `new_sync() -> Self`
2. `new_async() -> Result<Self>`
3. `new_async_with_rio(rio: Rio) -> Self`
4. `new_async_fallback_to_sync()`
5. `new()` - calls `new_async_fallback_to_sync()`

src/io/mod.rs
```diff
@@ -0,0 +1,13 @@
+#[cfg(target_os = "linux")]
+mod unix;
+#[cfg(target_os = "linux")]
+pub(crate) use unix::File;
+#[cfg(target_os = "linux")]
+pub use unix::IODriver;
+
+#[cfg(all(target_os = "macos", not(feature = "async-io-rio")))]
+mod linux;
```
ikopylov, line 9: `mod unix;`

src/io/mod.rs
```diff
@@ -0,0 +1,13 @@
...
+#[cfg(all(target_os = "macos", not(feature = "async-io-rio")))]
+pub(crate) use unix::File;
+#[cfg(all(target_os = "macos", not(feature = "async-io-rio")))]
+pub use unix::IODriver;
```
ikopylov, line 13: On this level declarations are identical, so no need to split them for `linux` and `macos`. It is better to use `#[cfg(target_family = "unix")]` and unify declarations

src/io/unix/mod.rs
```diff
@@ -0,0 +1,15 @@
...
+mod sync;
+#[cfg(not(feature = "async-io-rio"))]
+pub(crate) use sync::File;
+#[cfg(not(feature = "async-io-rio"))]
+pub use sync::IODriver;
+
+#[cfg(feature = "async-io-rio")]
+mod iouring;
+#[cfg(feature = "async-io-rio")]
+mod sync;
```
ikopylov, line 11: Please, add comment that `sync` needed as a base layer for `async`

src/io/unix/sync.rs
```diff
@@ -0,0 +1,197 @@
...
+use std::sync::atomic::AtomicU64;
+use std::{
+    os::unix::prelude::{AsRawFd, FileExt},
+    time::SystemTime,
+};
+use tokio::fs::OpenOptions;
+
+/// IO driver for file operations
+#[derive(Debug, Clone, Default)]
+pub struct IODriver;
```
ikopylov, line 16: `IoDriver` is better

src/io/unix/sync.rs
```diff
@@ -0,0 +1,197 @@
...
+
+impl IODriver {
```
ikopylov, line 18: Same as for async version: we need explicit constructors. Also it is good to remove `Default` implementation
1. `new_sync() -> Self`
2. `new() -> Self` - calls new_sync()

src/io/unix/sync.rs
```diff
@@ -0,0 +1,197 @@
...
+    async fn from_file(
+        path: impl AsRef<Path>,
+        setup: impl Fn(&mut OpenOptions) -> &mut OpenOptions,
+    ) -> IOResult<Self> {
+        let file = setup(&mut OpenOptions::new()).open(path.as_ref()).await?;
+
+        if Self::advisory_write_lock_file(file.as_raw_fd()) == LockAcquisitionResult::AlreadyLocked
+        {
+            error!("File {:?} is locked", path.as_ref());
+            panic!("File {:?} is locked", path.as_ref());
```
ikopylov, line 180: Do you remember, why the code is panic in this case? Looks like it is better to return `Error`

src/io/unix/sync.rs
```diff
@@ -0,0 +1,197 @@
...
+            .await
+            .expect("spawned blocking task failed")
+    }
+
+    fn inplace_sync_call<F, R>(f: F) -> R
+    where
+        F: FnOnce() -> R + Send + 'static,
+        R: Send + 'static,
+    {
+        match tokio::runtime::Handle::current().runtime_flavor() {
```
ikopylov, line 165: This is a bad approach. It is better to avoid `inplace_sync_call` at the caller level when runtime is `CurrentThread`. I suggest to add function `can_run_io_in_place(io_len: usize) -> bool` that checks both  `io_len <= MAX_SYNC_OPERATION_SIZE` and `runtime_flavor() != RuntimeFlavor::CurrentThread`

ikopylov, line 165: `current()` panics if there is no tokio runtime. I suggest to use `try_current()` instead

src/io/unix/sync.rs
```diff
@@ -0,0 +1,197 @@
...
+    pub(crate) async fn write_all_at(&self, offset: u64, buf: Bytes) -> IOResult<()> {
+        let file = self.no_lock_fd.clone();
+        if buf.len() <= MAX_SYNC_OPERATION_SIZE {
+            Self::inplace_sync_call(move || file.write_all_at(&buf, offset))
+        } else {
+            Self::background_sync_call(move || file.write_all_at(&buf, offset)).await
+        }
+    }
+
+    #[cfg(not(feature = "async-io-rio"))]
```
ikopylov, line 93: Don't see any reasons to exclude this function conditionally. It is better to mark it with `#[allow(dead_code)]`

src/io/unix/sync.rs
```diff
@@ -0,0 +1,197 @@
...
+    pub(crate) async fn read_all(&self) -> Result<BytesMut> {
+        self.read_exact_at_allocate(self.size().try_into()?, 0)
+            .await
+    }
+
+    #[cfg(not(feature = "async-io-rio"))]
```
ikopylov, line 99: Same here

src/io/unix/iouring.rs
```diff
@@ -0,0 +1,176 @@
...
+            debug!("blob file read at complited: {}", size);
+            Ok(buf)
+        } else {
+            self.sync.read_exact_at(buf, offset).await
+        }
+    }
+
+    pub(crate) async fn fsyncdata(&self) -> IOResult<()> {
+        if let Some(ref rio) = self.rio {
+            rio.fsync(&*self.sync.no_lock_fd).await
```
ikopylov, line 149: You can add `fn std_file_ref(&self) -> &StdFile` function to the sync `File` and use it here and in other places

src/io/unix/sync.rs
```diff
@@ -0,0 +1,197 @@
...
+                file.write_all_at(&first_buf, offset)?;
+                offset = offset + first_len;
+                file.write_all_at(&second_buf, offset)?;
+                Ok(())
+            })
+            .await
+        }
+    }
+
+    pub(crate) async fn write_all_at(&self, offset: u64, buf: Bytes) -> IOResult<()> {
```
ikopylov, line 84: Please, add `debug_assert!(offset + buf.len() <= self.size());`. We should be sure that this function will not change file size

src/io/unix/iouring.rs
```diff
@@ -0,0 +1,176 @@
...
+            offset = offset + first_len;
+            self.write_all_at_aio(offset, second_buf, rio).await
+        } else {
+            self.sync
+                .write_append_all_buffers(first_buf, second_buf)
+                .await
+        }
+    }
+
+    pub(crate) async fn write_all_at(&self, offset: u64, buf: Bytes) -> IOResult<()> {
```
ikopylov, line 83: Please, add `debug_assert!(offset + buf.len() <= self.size());`. We should be sure that this function will not change file size

src/io/unix/sync.rs
```diff
@@ -0,0 +1,197 @@
...
+#[derive(PartialEq, Eq)]
+enum LockAcquisitionResult {
+    Acquired,
+    AlreadyLocked,
+    Error(Errno),
+}
+
+impl File {
+    pub(crate) fn size(&self) -> u64 {
+        self.size.load(ORD)
```
ikopylov, line 48: In other places `Ordering::SeqCst` is used, so it is better to use it here too explicitly

src/io/unix/sync.rs
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    pub(crate) async fn write_append_all(&self, buf: Bytes) -> IOResult<()> {
+        let offset = self.size.fetch_add(buf.len() as u64, Ordering::SeqCst);
```
ikopylov, line 52: Add function `pub(super) fn file_size_append(&self, len: u64) -> u64` that executes exactly this line and use it here and in async `File`

src/io/unix/iouring.rs
```diff
@@ -0,0 +1,176 @@
...
+}
+
+impl File {
+    pub(crate) fn size(&self) -> u64 {
+        self.sync.size()
+    }
+
+    pub(crate) async fn write_append_all(&self, buf: Bytes) -> IOResult<()> {
+        if let Some(ref rio) = self.rio {
+            let offset = self.sync.size.fetch_add(buf.len() as u64, Ordering::SeqCst);
```
ikopylov, line 54: `file_size_append()` function from sync `File`

.github/workflows/build.yml
```diff
@@ -20,8 +19,13 @@ jobs:
...
+        with:
+          command: build
+          args: --all
+      - name: Build async rio
         uses: actions-rs/cargo@v1
         with:
           command: build
-          args: --all
\ No newline at end of file
+          args: --features async-io-rio
```
ikopylov, line 31: Same should be done for tests: https://github.com/qoollo/pearl/blob/master/.github/workflows/tests.yml#L27

First, run test with default features. Then add new step to run them with `async-io-rio` feature

src/storage/builder.rs
ikopylov: Actually, using `default` is bad for `IoDriver`. It is better to fallback to `IoDriver::new_sync()` if driver was not specified explicitly

src/tools/utils.rs
ikopylov: `IoDriver::new_sync()`

src/tools/validation.rs
ikopylov: `IoDriver::new_sync()`

src/error.rs
idruzhitskiy: Why is it no good? IMO this way we'll find out when it really becomes dead_code, so it's better to conditionally include it

src/storage/builder.rs
idruzhitskiy: So all iodrivers should have `new_sync`?

src/blob/index/benchmarks.rs
idruzhitskiy: Why? Default creator for async iodriver does not create `Rio`, and any default creator should be fast enough, if not const

src/io/unix/iouring.rs
idruzhitskiy: Why not single new, special for every iodriver, and single default, which should be minimal initialization constructor?

src/io/unix/sync.rs
idruzhitskiy: Because there is nothing we can do here, I guess

src/blob/index/benchmarks.rs
```diff
@@ -179,19 +176,14 @@ async fn benchmark_get_any() {
...
-    let ioring = ioring.ok();
     println!("Generating headers...");
     let headers = generate_headers(RECORDS_AMOUNT, KEY_MAPPER);
     let meta = generate_meta(META_SIZE);
 
     println!("Creating file index from headers...");
     let findex = FileIndexStruct::from_records(
         Path::new(FILEPATH),
-        ioring.clone(),
+        create_io_driver(),
```
ikopylov, line 186: It's probably better to follow the general style: create an `IoDriver` at the beginning, store it in a local field, and pass its clone here

src/blob/index/benchmarks.rs
```diff
@@ -233,19 +225,14 @@ async fn benchmark_get_all() {
...
```
ikopylov, line 235: Same here

src/blob/index/bptree/tests.rs
```diff
@@ -75,7 +79,7 @@ async fn serialize_deserialize_file() {
     let meta = vec![META_VALUE; META_SIZE];
     let findex = BPTreeFileIndex::<KeyType>::from_records(
         &Path::new("/tmp/bptree_index.b"),
-        None,
+        create_io_driver(),
```
ikopylov, line 82: Same in this whole file: create at the beginning of the function, store to field, pass here its clone

src/blob/index/bptree/tests.rs
```diff
@@ -65,6 +65,10 @@ impl Into<usize> for KeyType {
     }
 }
 
+fn create_io_driver() -> IoDriver {
+    IoDriver::new_sync()
```
ikopylov, line 69: I think it's better to create `IoDriver` here with `new()`, so that tests with `aync-io-rio` enabled use asynchronous I/O. Although previously tests were running in synchronous mode, now it is better to make it possible to run them with async mode too

src/io/mod.rs
```diff
@@ -0,0 +1,6 @@
+#[cfg(target_family = "unix")]
+mod unix;
+#[cfg(target_family = "unix")]
+pub(crate) use unix::File;
+#[cfg(target_family = "unix")]
+pub use unix::IoDriver;
```
ikopylov, line 6: It would be good to fail compilation with descriptive error for non-supported platforms with `compile_error!` macro (https://doc.rust-lang.org/std/macro.compile_error.html). 

```rust
#[cfg(not(target_family = "unix"))]
compile_error!("Specified target platform is not supported (only unix family supported)");
```

src/storage/builder.rs
```diff
@@ -29,10 +29,19 @@ use super::prelude::*;
...
     config: Config,
-    ioring: Option<Rio>,
+    iodriver: IoDriver,
+}
+
+impl Default for Builder {
+    fn default() -> Self {
+        Self {
+            config: Config::default(),
+            iodriver: IoDriver::new_sync(),
```
ikopylov, line 42: This assumes, that `IoDriver::new_sync()` is cheap. I'd prefer to declare field with option: `iodriver: Option<IoDriver>` and set it here to `None`. In this way there is a guaranty, that initialization of builder will always be cheap

tests/tests.rs
```diff
@@ -979,12 +990,6 @@ async fn test_blob_header_validation() {
         .max_data_in_blob(100_000)
         .set_filter_config(Default::default())
         .allow_duplicates();
-    let builder = if let Ok(ioring) = rio::new() {
```
ikopylov, line 982: You should set `IoDriver` explicitly here (with `IoDriver::new()`)

tests/tests.rs
```diff
@@ -246,7 +247,17 @@ async fn test_work_dir_lock() {
 
             let storage = res_one.unwrap();
             let result = waitpid(child, None);
+            #[cfg(not(target_os = "macos"))]
             assert_eq!(Ok(WaitStatus::Exited(child, 0)), result);
```
ikopylov, line 251: This test is incorrect. `fork` splits test engine as well, so the forked process can continue running other tests. Probably, that is the reason why the exit code is `0` (on panic it should not be 0)

ikopylov: Please, update to use `rusty-fork` crate (https://docs.rs/rusty-fork/0.3.0/rusty_fork/fn.fork.html)

src/io/unix/sync.rs
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    pub(crate) fn created_at(&self) -> Result<SystemTime> {
+        let metadata = self.no_lock_fd.metadata()?;
+        Ok(metadata.created().unwrap_or(SystemTime::now()))
+    }
+
+    #[cfg(feature = "async-io-rio")]
+    pub(super) fn std_file_ref(&self) -> &StdFile {
+        &*self.no_lock_fd
```
ikopylov, line 144:
```suggestion
        self.no_lock_fd.as_ref()
```

src/io/unix/sync.rs
```diff
@@ -0,0 +1,222 @@
...
+        F: FnOnce() -> R + Send + 'static,
+        R: Send + 'static,
+    {
+        tokio::task::block_in_place(move || f())
+    }
+
+    fn can_run_inplace(len: u64) -> bool {
+        use tokio::runtime::{Handle, RuntimeFlavor};
+        Handle::current().runtime_flavor() != RuntimeFlavor::CurrentThread
+            && len <= MAX_SYNC_OPERATION_SIZE as u64
```
ikopylov, line 193: I suggest to compare `len` first, because that check is faster

src/io/unix/sync.rs
ikopylov: Ok, lets keep this as is. I filled the issue for the future: https://github.com/qoollo/pearl/issues/253

tests/tests.rs
ikopylov: Actually, lets do this in a separate request. I added a new issue: https://github.com/qoollo/pearl/issues/254 and assigned it to you

src/storage/builder.rs
```diff
@@ -29,10 +29,19 @@ use super::prelude::*;
...
 /// [`key_size`]: struct.Builder.html#method.key_size
-#[derive(Default, Debug)]
+#[derive(Debug)]
 pub struct Builder {
     config: Config,
-    ioring: Option<Rio>,
+    iodriver: Option<IoDriver>,
+}
+
+impl Default for Builder {
```
ikopylov, line 38: In this form, `Default` can be implemented with `derive`

src/storage/builder.rs
```diff
@@ -183,6 +188,12 @@ impl Builder {
         self
     }
 
+    /// Set io driver to be used during file io operations
```
ikopylov, line 191:
```suggestion
    /// Set io driver to be used during file io operations. If the driver is not set, then `sync` is used as the default
```

src/io/unix/iouring.rs
```diff
@@ -0,0 +1,194 @@
...
+    pub(crate) async fn read_exact_at_allocate(
+        &self,
+        size: usize,
+        offset: u64,
+    ) -> Result<BytesMut> {
+        let buf = BytesMut::zeroed(size);
+        self.read_exact_at(buf, offset).await
+    }
+
+    pub(crate) async fn read_exact_at(&self, mut buf: BytesMut, offset: u64) -> Result<BytesMut> {
```
ikopylov, line 123: I think, it is better to split this function into 2:
1. `read_exact_at_aio` contains core logic for read with `Rio`
2. `read_exact_at` checks that `self.rio` is `Some(..)` and calls either `read_exact_at_aio(..)` or `self.sync.read_exact_at(..)`

src/io/unix/iouring.rs
```diff
@@ -0,0 +1,198 @@
...
```
ikopylov, line 123: There is a warning about `mut`

```suggestion
    pub(crate) async fn read_exact_at(&self, buf: BytesMut, offset: u64) -> Result<BytesMut> {
```

src/io/unix/iouring.rs
```diff
@@ -0,0 +1,198 @@
...
+        debug!("blob file read at");
+        if buf.is_empty() {
+            warn!("file read_at empty buf");
```
ikopylov, line 126: Actually, it is better to remove this warn, because in general it is ok to read into an empty buffer. A `debug` log above is also useless

ikopylov (approved):
LGTM

PR #255, opened 2023-03-30
254-better-fork

### LINKED ISSUE #254: Remove raw `fork` from tests and use `rusty-fork` instead
https://github.com/qoollo/pearl/blob/v0.18.0/tests/tests.rs#L239

`fork` withing the tests is incorrect, since the test engine is also forked. We should use a different approach. For example, the crate `rusty-fork` (https://crates.io/crates/rusty-fork) runs the same test with additional parameters. It's worth to use it

idruzhitskiy:
Closes #254


Cargo.toml
```diff
@@ -50,6 +50,7 @@ async-lock = "2.6"
 # Benchmark only dependencies
 clap = { version = "3.2", optional = true }
 env_logger = { version = "0.9", optional = true }
+rusty-fork = "0.3.0"
```
ikopylov, line 53: This dependency should be within `dev-dependencies`, because it is only needed for tests

tests/tests.rs
```diff
@@ -229,50 +229,49 @@ async fn test_on_disk_index() -> Result<()> {
...
+    use std::sync::Arc;
+    use tokio::runtime::Builder;
 
+    let path = Arc::new(common::init("work_dir_lock"));
+    let child_path = path.clone();
     // We need separate processes because locks do not work within the same process
-    match unsafe { fork() } {
-        Ok(ForkResult::Parent { child }) => {
+    fork(
+        "test_work_dir_lock",
```
ikopylov, line 242: It would be good to not hardcode the function name. I found this crate: https://docs.rs/function_name/latest/function_name/ . It adds a macro to get the name of the function

ikopylov: According to the documentation, you should wrap the function name with `rusty_fork_test_name!` macro (https://docs.rs/rusty-fork/latest/rusty_fork/macro.rusty_fork_test_name.html)

tests/tests.rs
```diff
@@ -229,50 +229,49 @@ async fn test_on_disk_index() -> Result<()> {
...
-        Ok(ForkResult::Child) => {
-            sleep(Duration::from_secs(1)).await;
-            let storage_two = common::create_test_storage(&path, 1_000_000);
-            let _ = storage_two.await;
+        },
+        move || {
+            let runtime = Builder::new_current_thread()
+                .build()
+                .expect("failed to create runtime in child");
+            std::thread::sleep(Duration::from_secs(1));
```
ikopylov, line 268: This is not safe, because the main process may not initialize in 1 second. I propose to create a special file (e.g. `main_init_ready.special`) in the main process right after its initialization complete, and here, in a loop with shorter `sleep`, wait for its appearance. Loop iteration count should be limited to avoid infinite waiting in case of some error

tests/tests.rs
```diff
@@ -229,50 +229,71 @@ async fn test_on_disk_index() -> Result<()> {
...
     // We need separate processes because locks do not work within the same process
-    match unsafe { fork() } {
-        Ok(ForkResult::Parent { child }) => {
+    fork(
+        rusty_fork_test_name!(test_work_dir_lock),
+        rusty_fork_id!(),
+        |_| {},
+        move |c, _| {
+            create_dir_all(path.as_ref()).expect("failed to create path to init");
+            write(init_file, vec![]).expect("failed to create init file");
```
ikopylov, line 254: This file is created too early. It should be created after the storage initialization completed (after the line `let res_one = runtime.block_on(storage_one);`)

tests/tests.rs
```diff
@@ -229,50 +229,71 @@ async fn test_on_disk_index() -> Result<()> {
...
-                )),
-                result
-            );
+            let exit = c
+                .wait_timeout(Duration::from_secs(MAX_PARENT_WAIT_SECONDS))
+                .expect("failed to wait for child");
+            if let Some(exit) = exit {
+                assert!(!exit.success());
+            } else {
+                c.kill().expect("failed to kill child process");
```
ikopylov, line 271: The test should fail in this case

tests/tests.rs
```diff
@@ -229,50 +229,71 @@ async fn test_on_disk_index() -> Result<()> {
...
+
+    let path = Arc::new(common::init("work_dir_lock"));
+    let init_file = Path::new(path.as_ref()).join("main_init_ready.special");
+    let child_init_file = init_file.clone();
+    let child_path = path.clone();
     // We need separate processes because locks do not work within the same process
-    match unsafe { fork() } {
-        Ok(ForkResult::Parent { child }) => {
+    fork(
+        rusty_fork_test_name!(test_work_dir_lock),
```
ikopylov, line 249: This macro does not check, that the actual function name matches the passed identifier. That's why I suggested to use a `function_name` crate: https://docs.rs/function_name/latest/function_name/

tests/tests.rs
```diff
@@ -229,50 +229,71 @@ async fn test_on_disk_index() -> Result<()> {
...
+            let mut attempt = 0;
+            while !read(&child_init_file).is_ok() {
+                assert!(attempt < MAX_CHILD_WAIT_COUNT);
+                attempt = attempt + 1;
+
+                std::thread::sleep(Duration::from_millis(200));
+            }
+            let storage_two = common::create_test_storage(child_path.as_ref(), 1_000_000);
+            let _ = runtime.block_on(storage_two);
             unreachable!("Second storage process must panic")
```
ikopylov, line 293: Here is the problem: if `create_test_storage` not panicked, then the panic happens here. In that case parent process decided that the test passed, which is not correct.
I found this function in rust: https://doc.rust-lang.org/std/panic/fn.catch_unwind.html . It allows to catch the panic. So I suggest to wrap `runtime.block_on(storage_two)` with `catch_unwind` and return success if the code panicked and error when it is not

tests/tests.rs
```diff
@@ -229,50 +230,80 @@ async fn test_on_disk_index() -> Result<()> {
...
+
+    let path = Arc::new(common::init(function_name!()));
+    let init_file = Path::new(path.as_ref()).join("main_init_ready.special");
+    let child_init_file = init_file.clone();
+    let child_path = path.clone();
     // We need separate processes because locks do not work within the same process
-    match unsafe { fork() } {
-        Ok(ForkResult::Parent { child }) => {
+    fork(
+        function_name!(),
```
ikopylov, line 252: You should keep `rusty_fork_test_name!()` too, because according to the source code it adds the module name (https://docs.rs/rusty-fork/latest/src/rusty_fork/fork_test.rs.html#121-126):
 ```suggestion
        rusty_fork_test_name!(function_name!()),
```

File: tests/tests.rs [hunk 9]
```diff
@@ -229,50 +230,80 @@ async fn test_on_disk_index() -> Result<()> {
...
+
+    let path = Arc::new(common::init(function_name!()));
+    let init_file = Path::new(path.as_ref()).join("main_init_ready.special");
+    let child_init_file = init_file.clone();
+    let child_path = path.clone();
     // We need separate processes because locks do not work within the same process
-    match unsafe { fork() } {
-        Ok(ForkResult::Parent { child }) => {
+    fork(
+       rusty_fork_test_name!(function_name!()),
```
ikopylov, line 252: Build error: `cannot find macro 'rusty_fork_test_name' in this scope`

tests/tests.rs
```diff
@@ -229,50 +230,80 @@ async fn test_on_disk_index() -> Result<()> {
...
-async fn test_work_dir_lock() {
-    use nix::sys::wait::{waitpid, WaitStatus};
-    use nix::unistd::{fork, ForkResult};
-
-    let path = common::init("work_dir_lock");
-
+#[test]
+#[named]
+fn test_work_dir_lock() {
+    use rusty_fork::{fork, rusty_fork_id};
```
ikopylov, line 236:
```suggestion
    use rusty_fork::{fork, rusty_fork_id, rusty_fork_test_name};
```

tests/tests.rs
```diff
@@ -229,50 +230,80 @@ async fn test_on_disk_index() -> Result<()> {
...
+
+    let path = Arc::new(common::init(function_name!()));
+    let init_file = Path::new(path.as_ref()).join("main_init_ready.special");
+    let child_init_file = init_file.clone();
+    let child_path = path.clone();
     // We need separate processes because locks do not work within the same process
-    match unsafe { fork() } {
-        Ok(ForkResult::Parent { child }) => {
+    fork(
+       rusty_fork_test_name!(function_name!()),
```
ikopylov, line 252: Build error here: `no rules expected this token in macro call`.
Unfortunately, `rusty_fork_test_name!` expects identifier as parameter, but `function_name!` is another macro.
In this case, it is better to reject `function_name!` macro usage and return to the previous implementation, where function name hardcoded here directly: `rusty_fork_test_name!(test_work_dir_lock)`

tests/tests.rs
```diff
@@ -229,50 +229,79 @@ async fn test_on_disk_index() -> Result<()> {
...
+
+    let path = Arc::new(common::init("test_work_dir_lock"));
+    let init_file = Path::new(path.as_ref()).join("main_init_ready.special");
+    let child_init_file = init_file.clone();
+    let child_path = path.clone();
     // We need separate processes because locks do not work within the same process
-    match unsafe { fork() } {
-        Ok(ForkResult::Parent { child }) => {
+    fork(
+        rusty_fork_test_name!(test_work_dir_lock),
```
ikopylov, line 250: According to my tests, error in the function name does not lead to the test failure. I suggest to add explicit check, that child process was started. This can be done by creating a file on the disk right at the beginning of the child process (the same way, like you create file to wait for initialization complete), and check that that file exists at the end of parent process

PR #256, opened 2023-04-26
249 lock-free write

### LINKED ISSUE #249: Get rid of `mut` modifier in the `index::push` function
This can be done by wrapping the internal in-memory index with sync RwLock. This way we can avoid write lock acquisition at all on write path, which should improve performance in concurrent read/write scenarios. Inner RwLock will be taken only for a short period of time (much shorter than RwLock on Blob)

https://github.com/qoollo/pearl/blob/master/src/blob/index/core.rs#L259
https://github.com/qoollo/pearl/blob/master/src/blob/core.rs#L283

idruzhitskiy:
Closes #249

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Pearl changelog
 
 #### Changed
 - Remove hard dependency on io uring (#230)
+- Remove upgradable lock from write path (#249)
 
 #### Fixed
 - Fix fork in test duplicating test environment (#254)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ Pearl changelog
 
 #### Changed
 - Remove hard dependency on io uring (#230)
-- Remove upgradable lock from write path (#249)
+- BLOB-wide exclusive lock removed from the write path, instead the short lock on the in-memory index added (#249)
 
 #### Fixed
 - Fix fork in test duplicating test environment (#254)
--- a/src/blob/core.rs
+++ b/src/blob/core.rs
@@ -272,7 +272,7 @@ where
     ) -> Result<()> {
         let write_result = record.write_to_file(&blob.file).await?;
         header.set_offset_checksum(write_result.blob_offset(), write_result.header_checksum());
-        let mut blob = RwLockUpgradableReadGuard::upgrade(blob).await;
+        let blob = RwLockUpgradableReadGuard::upgrade(blob).await;
         blob.index.push(key, header)?;
         Ok(())
     }
--- a/src/blob/core.rs
+++ b/src/blob/core.rs
@@ -247,9 +247,10 @@ where
 
     pub(crate) async fn write(blob: &ASRwLock<Self>, key: &K, record: Record) -> Result<()> {
         debug!("blob write");
-        // Only one upgradable_read lock is allowed at a time
         let (partially_serialized, mut header) = record.to_partially_serialized_and_header()?;
-        let blob = blob.read().await;
+        // Only one upgradable_read lock is allowed at a time. This is critical because we want to
+        // be sure that only one write operation is running at a time
+        let blob = blob.upgradable_read().await;
         let write_result = partially_serialized.write_to_file(&blob.file).await?;
         header.set_offset_checksum(write_result.blob_offset(), write_result.header_checksum());
         blob.index.push(key, header)?;
--- a/src/blob/core.rs
+++ b/src/blob/core.rs
@@ -29,7 +29,6 @@ where
     file: File,
     created_at: SystemTime,
     validate_data_during_index_regen: bool,
-    write_sem: Semaphore,
 }
 
 impl<K> Blob<K>
@@ -62,7 +61,6 @@ where
             file,
             created_at: SystemTime::now(),
             validate_data_during_index_regen,
-            write_sem: Semaphore::new(1),
         };
         blob.write_header().await?;
         Ok(blob)
@@ -182,7 +180,6 @@ where
             index,
             created_at,
             validate_data_during_index_regen,
-            write_sem: Semaphore::new(1),
         };
         trace!("call update index");
         if is_index_corrupted || size as u64 > header_size {
@@ -249,11 +246,16 @@ where
         // @TODO implement
     }
 
-    pub(crate) async fn write(blob: &RwLock<Self>, key: &K, record: Record) -> Result<()> {
+    pub(crate) async fn write(
+        blob: &RwLock<Self>,
+        write_sem: &Semaphore,
+        key: &K,
+        record: Record,
+    ) -> Result<()> {
         debug!("blob write");
         let (partially_serialized, mut header) = record.to_partially_serialized_and_header()?;
+        let _ = write_sem.acquire().await?;
         let blob = blob.read().await;
-        let _ = blob.write_sem.acquire().await?;
         let write_result = partially_serialized.write_to_file(&blob.file).await?;
         header.set_offset_checksum(write_result.blob_offset(), write_result.header_checksum());
         blob.index.push(key, header)?;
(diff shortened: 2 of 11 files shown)


src/blob/core.rs
```diff
@@ -251,8 +250,8 @@ where
         debug!("blob write");
         // Only one upgradable_read lock is allowed at a time
         let (partially_serialized, header) = record.to_partially_serialized_and_header()?;
-        let blob = blob.upgradable_read().await;
-        Self::write_locked(blob, key, partially_serialized, header).await
+        let blob = blob.read().await;
```
ikopylov, line 253: This change is questionable, because it allows multiple simultaneous writes into a single file. We should estimate pros and cons of this change

src/blob/core.rs
```diff
@@ -251,8 +250,8 @@ where
...
+        Self::write_with_lock(blob, key, partially_serialized, header).await
```
ikopylov, line 254: I think the code from `write_with_lock` can be moved in here and `write_with_lock` can be removed

src/blob/index/core.rs
```diff
@@ -55,20 +55,20 @@ where
...
+    pub(crate) records_count: AtomicUsize,
+    pub(crate) records_allocated: AtomicUsize,
 }
 
 pub type InMemoryIndex<K> = BTreeMap<K, Vec<RecordHeader>>;
 
 #[derive(Debug, Clone)]
 pub(crate) enum State<FileIndex, K> {
-    InMemory(InMemoryIndex<K>),
+    InMemory(Arc<SRwLock<Option<InMemoryIndex<K>>>>),
```
ikopylov, line 71: `Arc` and `Option` are not needed. They only give overhead. `Clone` on the `State` can be removed

src/blob/index/core.rs
```diff
@@ -55,20 +55,20 @@ where
 
 #[derive(Debug, Default)] // Default can be used to initialize structure with 0
 pub(crate) struct MemoryAttrs {
```
ikopylov, line 57: Please, recheck where `MemoryAttrs` is used. I see that it is only used in this module to calculate occupied memory by `InMemory` index. If it is true, then it is better to move `mem: Option<MemoryAttrs>` field from `IndexStruct` into `State::InMemory` and place it under the lock. In this way, there is not need to wrap it with `Option` + fields can be a of regular `usize` types.

Also, if my observation is correct, then `key_size`, `btree_entry_size` and `record_header_size` can be converted into `const` by making `MemoryAttrs` a generic structure, parametrized by key type

src/blob/core.rs
ikopylov: Can you run benchmark with simultaneous writes from multiple threads into a single blob to compare `read()` vs `upgradable_read()` locks?
I expect that when the write goes to the disk cache, the `read()` lock will be faster, but when the disk cache is full, then `upgradable_read()` will become faster. 

`upgradable_read()` implies that only one thread is writing at a time, and this is generally safer in case of emergencies

src/blob/index/core.rs
idruzhitskiy: Indeed, `mem::take` can be used!

src/blob/core.rs
idruzhitskiy: I've run some tests, using `upgradable_read` compared to master slightly improves read performance, and write performance stays the same. Using `read` compared to `upgradable_read` improves write performance without any noticable degradation in other aspects.

src/blob/index/core.rs
idruzhitskiy: Done, now part of `MemoryAttrs` is const, also removed unnecessary spread of functions related to this struct from other modules

src/blob/index/core.rs
```diff
@@ -54,21 +53,82 @@ where
...
+
+impl<K> MemoryAttrs<K>
+where
+    for<'a> K: Key<'a>,
+{
+    const fn key_size(&self) -> usize {
+        K::LEN as usize
+    }
+
+    const fn bptree_entry_size(&self) -> usize {
```
ikopylov, line 70: According to the documentation `const fn` works as a regular function, but additionally they can be used to initialize a constant. That means, that it is up to the optimizer to fold all this function into a single constant. I've checked, that on debug level Rust compiler emits a full calculation logic. Thus I suggest to declare a regular constant here instead of `const fn`:

```
const KEY_SIZE: usize = K::LEN as usize;
const BPTREE_ENTRY_SIZE: usize = size_of::<Vec<u8>>() + KEY_SIZE + size_of::<Vec<RecordHeader>>();
const RECORD_HEADER_SIZE: usize = size_of::<RecordHeader>() + KEY_SIZE;
```

You can keep these functions for simplicity, but they should return a declared constants and be attributed with `#[inline]`

src/blob/index/core.rs
```diff
@@ -54,21 +53,82 @@ where
...
+            records_count,
+            records_allocated,
+            ..
+        } = &mem;
+        let len = self.headers.len();
+        trace!("record_header_size: {}, records_allocated: {}, data.len(): {}, entry_size (key + vec): {}",
+                mem.record_header_size(), records_allocated, len, mem.bptree_entry_size());
+        // last minus is neccessary, because allocated but not initialized record
+        // headers don't have key allocated on heap
+        mem.record_header_size() * records_allocated + len * mem.bptree_entry_size()
```
ikopylov, line 115: Looks like, we count less, than the actual size of BTreeMap in memory here. I will add a separate issue for this problem

ikopylov: https://github.com/qoollo/pearl/issues/266

src/blob/index/core.rs
idruzhitskiy: I don't think functions are needed here, replaced with `const`s

src/blob/index/core.rs
```diff
@@ -54,21 +53,74 @@ where
...
+        MemoryAttrs::<K>::RECORD_HEADER_SIZE * records_allocated
+            + len * MemoryAttrs::<K>::BPTREE_ENTRY_SIZE
+            - (records_allocated - records_count) * MemoryAttrs::<K>::KEY_SIZE
+    }
+
+    fn records_count(&self) -> usize {
+        self.mem.records_count
+    }
+
+    fn register_record_allocation(&mut self, allocated: usize) {
```
ikopylov, line 115: Rename `allocated` to `records_allocated`

src/blob/index/core.rs
```diff
@@ -256,31 +295,26 @@ where
...
+                let mut data = headers.write().expect("rwlock");
                 debug!("blob index simple push bloom filter add");
                 self.filter.add(key);
                 debug!("blob index simple push key: {:?}", h.key());
-                let mem = self
-                    .mem
-                    .as_mut()
-                    .expect("No memory info in `InMemory` State");
-                if let Some(v) = headers.get_mut(key) {
+                let allocated;
```
ikopylov, line 306: Same here: rename `allocated` to `records_allocated`

src/blob/core.rs
ikopylov: Please, replace `RwLock` from `async-lock` crate to `RwLock` from `tokio`, because there is no need to use read lock upgrade feature anymore. To ensure single writer execution at a time, you can add `Semaphore` inside `Blob` structure.

Additionally, please run the test with and without `Semaphore`, running an additional `bobp` instance that writes large (~100Mb) records

ikopylov: According to my tests and investigations, it is better to return back to the `RwLock` from `async-lock` crate (revert last 2 commits) and use `upgradable_read`.
Additionally, we should add a comment: `Only one upgradable_read lock is allowed at a time. This is critical because we want to be sure that only one write operation is running at a time.`

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ Pearl changelog
 
 #### Changed
 - Remove hard dependency on io uring (#230)
+- Remove upgradable lock from write path (#249)
```
ikopylov, line 11:
```suggestion
- BLOB-wide exclusive lock removed from the write path, instead the short lock on the in-memory index added (#249)
```

ikopylov (approved):
LGTM

PR #258, opened 2023-04-27
Add limited Windows support, fix bug in ObserverWorker, improve Arc placement and visibility levels

### LINKED ISSUE #185: Review and update `Arc` placement in `Storage`
Looks like only `inner` should be wrapped in `Arc` (https://github.com/qoollo/pearl/blob/master/src/storage/core.rs#L43), because it is actually passed outside. All inner fields like `safe`, `next_blob_id`, `blobs` does not pass outside (from the first glance), so it looks like `Arc` on them can be safely removed

### LINKED ISSUE #186: Review and update visibility levels in `Storage`
Too many unnecessary `pub(crate)` usage (https://github.com/qoollo/pearl/blob/master/src/storage/core.rs).
Looks like `ioring` should be totally private (https://github.com/qoollo/pearl/blob/master/src/storage/core.rs#L55)
`config` and `safe` can have readonly access, so they should be exposed through the getter and declared as private (https://github.com/qoollo/pearl/blob/master/src/storage/core.rs#L52).
Same for fields of `Safe`: they should not be available for updated outside (https://github.com/qoollo/pearl/blob/master/src/storage/core.rs#L63). From the architecture point of view, `Safe` should not visible at all, because it is an integral part of `Inner<K>` protected by lock (this is discussable).

### LINKED ISSUE #257: Add limited Windows support
Pearl should build and work on Windows, tests should pass. Performance is not important

### LINKED ISSUE #259: ObserverWorker never stops due to loopback sender
`loopback_sender` prevents stopping of `ObserverWorker`. That leads to memory leak

ikopylov:
Closes #257 
Closes #259
Closes #185
Closes #186

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,16 +4,19 @@ Pearl changelog
 
 ## [Unreleased]
 #### Added
-
+- Limited Windows support added (#257)
 
 #### Changed
 - Remove hard dependency on io uring (#230)
+- Update offset only when write is imminent to prevent data corruption due to async Future cancellation (#228)
+- Improved Arc placement in Storage (#185)
+- Improved visibility levels in Storage (#186)
 
 #### Fixed
 - Fix fork in test duplicating test environment (#254)
+- Bug in ObserverWorker that prevents its stop and leads to memory leak (#259)
 
 #### Updated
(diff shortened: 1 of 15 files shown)


ikopylov:
Was split into https://github.com/qoollo/pearl/pull/264 and https://github.com/qoollo/pearl/pull/265

PR #264, opened 2023-05-02
[257] Add limited Windows support

### LINKED ISSUE #257: Add limited Windows support

ikopylov:
Closes #257

### DIFF
--- a/src/blob/index/bptree/tests.rs
+++ b/src/blob/index/bptree/tests.rs
@@ -68,7 +68,7 @@ impl Into<usize> for KeyType {
 fn get_test_dir(dir_name: &str) -> PathBuf {
     let result = std::env::temp_dir()
         .join("pearl_bptree_test")
-        .join((std::time::UNIX_EPOCH.elapsed().unwrap().as_secs() % 1_563_100_000).to_string())
+        .join(std::time::UNIX_EPOCH.elapsed().unwrap().as_secs().to_string())
         .join(dir_name);
 
     std::fs::create_dir_all(&result).expect("Directory created");
--- a/src/blob/index/bptree/tests.rs
+++ b/src/blob/index/bptree/tests.rs
@@ -65,12 +65,29 @@ impl Into<usize> for KeyType {
     }
 }
 
+fn get_test_dir(dir_name: &str) -> PathBuf {
+    let result = std::env::temp_dir()
+        .join("pearl_bptree_test")
+        .join((std::time::UNIX_EPOCH.elapsed().unwrap().as_secs() % 1_563_100_000).to_string())
+        .join(dir_name);
+
+    std::fs::create_dir_all(&result).expect("Directory created");
+
+    return result;
+}
+
+fn clean(index: BPTreeFileIndex<KeyType>, path: impl AsRef<Path>) -> Result<()> {
+    std::mem::drop(index);
+    std::fs::remove_dir_all(path).map_err(Into::into)
+}
+
 fn create_io_driver() -> IoDriver {
     IoDriver::new()
 }
 
 #[tokio::test(flavor = "multi_thread")]
 async fn serialize_deserialize_file() {
+    let test_dir = get_test_dir("serialize_deserialize_file");
     let mut inmem = InMemoryIndex::<KeyType>::new();
     (0..10000).map(|i| i.into()).for_each(|key: KeyType| {
         let rh = RecordHeader::new(key.to_vec(), 1, 1, 1);
@@ -79,7 +96,7 @@ async fn serialize_deserialize_file() {
     let meta = vec![META_VALUE; META_SIZE];
     let iodriver = create_io_driver();
     let findex = BPTreeFileIndex::<KeyType>::from_records(
-        &Path::new("/tmp/bptree_index.b"),
+        &test_dir.join("bptree_index.b"),
         iodriver,
         &inmem,
         meta,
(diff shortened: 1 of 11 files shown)


src/blob/index/bptree/tests.rs
```diff
@@ -65,12 +65,29 @@ impl Into<usize> for KeyType {
...
+        .join("pearl_bptree_test")
+        .join((std::time::UNIX_EPOCH.elapsed().unwrap().as_secs() % 1_563_100_000).to_string())
+        .join(dir_name);
+
+    std::fs::create_dir_all(&result).expect("Directory created");
+
+    return result;
+}
+
+fn clean(index: BPTreeFileIndex<KeyType>, path: impl AsRef<Path>) -> Result<()> {
```
idruzhitskiy, line 79: Since every occurence of `clean` is accompanied by `expect` it's probably better to remove return value from this function and move `expect` inside.

src/io/unix/iouring.rs
```diff
@@ -197,3 +197,37 @@ impl File {
         Ok(Self { sync, rio })
     }
 }
+
+
+/// Trait implementation to track that all required function are implemented
+#[async_trait::async_trait]
+impl super::super::FileTrait for File {
```
idruzhitskiy, line 204: Why not move implementations into trait?

src/io/windows/sync.rs
```diff
@@ -0,0 +1,238 @@
...
+        let file_inner = self.inner.clone();
+        Self::background_sync_call(move || file_inner.std_file.sync_all()).await
+    }
+
+    pub(crate) fn created_at(&self) -> IOResult<SystemTime> {
+        let metadata = self.inner.std_file.metadata()?;
+        Ok(metadata.created().unwrap_or(SystemTime::now()))
+    }
+
+    async fn background_sync_call<F, R>(f: F) -> R
```
idruzhitskiy, line 125: This function and also inplace call can be extracted into separate mod available to all io drivers. Also threshold can be made generic const so each platform will have its own compile-time generated instance of async-sync worker.
This probably can be made into separate ticket

src/io/windows/sync.rs
```diff
@@ -0,0 +1,238 @@
...
+    fn read_exact_at(&self, mut buf: &mut [u8], mut offset: u64) -> IOResult<()> {
+        while !buf.is_empty() {
+            match self.std_file.seek_read(buf, offset) {
+                Ok(0) => break,
+                Ok(n) => {
+                    let tmp = buf;
+                    buf = &mut tmp[n..];
+                    offset += n as u64;
+                }
+                Err(ref e) if e.kind() == std::io::ErrorKind::Interrupted => {}
```
idruzhitskiy, line 177: `std::io::{Error, ErrorKind}` should be imported

src/io/windows/sync.rs
```diff
@@ -0,0 +1,238 @@
...
+                Err(e) => return Err(e),
+            }
+        }
+        if !buf.is_empty() {
+            Err(std::io::Error::new(std::io::ErrorKind::UnexpectedEof, "failed to fill whole buffer"))
```
idruzhitskiy, line 182: There is only one possibility for this if to work - `Ok(0)` branch, right? Why not move err into this branch?

idruzhitskiy: Or change write to look the same

src/blob/index/bptree/tests.rs
```diff
@@ -65,12 +65,29 @@ impl Into<usize> for KeyType {
     }
 }
 
+fn get_test_dir(dir_name: &str) -> PathBuf {
+    let result = std::env::temp_dir()
+        .join("pearl_bptree_test")
+        .join((std::time::UNIX_EPOCH.elapsed().unwrap().as_secs() % 1_563_100_000).to_string())
```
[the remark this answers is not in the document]
ikopylov: This is not a constant. It is the time in seconds elapsed from `UNIX_EPOCH`. It should be calculated for each test run to avoid running on top of previous run results

src/io/unix/iouring.rs
ikopylov: I've added an additional comment on top of each implementation explaining the reason why it should not contain actual function body inside

src/io/windows/sync.rs
ikopylov: I don't think it's needed now. This function is trivial, and this is only its second copy.

I didn't add `inplace_sync_call` to the Windows implementation at all, because the Windows implementation is unlikely to be used in production, and I don't want to spend time running tests to find the best threshold

src/io/windows/sync.rs
ikopylov: This is common type names, so I've imported them with `std::io::{Error as IOError, ErrorKind as IOErrorKind}`

src/io/windows/sync.rs
ikopylov: Ok, I've synced the implementation of `read_exact_at` and `write_all_at`

src/io/unix/iouring.rs
ikopylov: https://blog.rust-lang.org/inside-rust/2023/05/03/stabilizing-async-fn-in-trait.html

src/blob/index/bptree/tests.rs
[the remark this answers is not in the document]
idruzhitskiy: I mean `1_563_100_000`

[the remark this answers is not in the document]
ikopylov: That is not important constant. I copied initialization from `tests/common.rs`. My guess, is that it was chosen to reduce the length of the string representation of the number (current UNIX timestamp is `1683232306`).  I think I'll just remove it

idruzhitskiy:
LGTM

PR #265, opened 2023-05-02
Observer worker never stops bug fix, improve Arc placement and visibility levels

### LINKED ISSUE #185: Review and update `Arc` placement in `Storage`

### LINKED ISSUE #186: Review and update visibility levels in `Storage`

### LINKED ISSUE #259: ObserverWorker never stops due to loopback sender

ikopylov:
Closes #259 
Closes #185 
Closes #186

### DIFF
--- a/src/storage/observer_worker.rs
+++ b/src/storage/observer_worker.rs
@@ -88,7 +88,7 @@ where
 
     async fn tick_with_deadline(&mut self, deadline: Instant) -> Result<TickResult> {
         // Extend deadline a little bit to guaranty, that process_defered will detect exceeding
-        let deadline = deadline + Duration::from_millis(10);
+        let deadline = deadline + Duration::from_millis(1);
         match timeout_at(deadline, self.receiver.recv()).await {
             Ok(Some(msg)) => {
                 self.process_msg(msg).await?;
--- a/src/storage/observer_worker.rs
+++ b/src/storage/observer_worker.rs
@@ -17,8 +17,6 @@ where
     dump_sem: Arc<Semaphore>,
     async_oplock: Arc<Mutex<()>>,
     deferred: Arc<Mutex<Option<DeferredEventData>>>,
-    deferred_min_duration: Duration,
-    deferred_max_duration: Duration,
 }
 
 struct DeferredEventData {
@@ -37,8 +35,6 @@ where
         dump_sem: Arc<Semaphore>,
         async_oplock: Arc<Mutex<()>>,
     ) -> Self {
-        let deferred_min_duration = inner.config().deferred_min_time().clone();
-        let deferred_max_duration = inner.config().deferred_max_time().clone();
         Self {
             inner,
             receiver,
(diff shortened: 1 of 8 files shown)


src/storage/observer_worker.rs
```diff
@@ -1,147 +1,218 @@
...
+                self.process_msg(msg).await?;
+                Ok(TickResult::Continue)
+            },
+            None => Ok(TickResult::Stop)
+        }
+    }
+
+    async fn tick_with_deadline(&mut self, deadline: Instant) -> Result<TickResult> {
+        // Extend deadline a little bit to guaranty, that process_defered will detect exceeding
+        let deadline = deadline + Duration::from_millis(10);
```
idruzhitskiy, line 91: Shouldn't this duration depend on config's defered min/max time?

ikopylov: Yes, the `deadline` is updated in `deffer_blob_indexes_dump` and in `process_deferred_blob_index_dump` according to  deferred min/max time from config.

The idea of this change is to update the `deadline`, every time `DeferredDumpBlobIndexes` message income and on every loop, when deadline is not `None` runs `recv` with additional timeout. If timeout triggered, run processing of deferred index dump and reset `deadline` value to `None.

idruzhitskiy: But `Duration::from_millis(10)` does not depend on config values, and therefore can be too large/too small in some situations, if I understood correctly

ikopylov: The idea is to set the duration a little longer than the original one, so that the check in `process_deferred_blob_index_dump` is guaranteed to be successful.

> But Duration::from_millis(10) does not depend on config values, and therefore can be too large/too small in some situations, if I understood correctly

The original concept of deferred execution is that the processing will happen at some time later. If it happens an additional 10ms later, this is not critical. The initial values in the config are measured in minutes.

Actually, I think, it would be good to reduce `10ms` to `1ms`

idruzhitskiy:
LGTM

PR #267, opened 2023-05-03
262 include bptree root mem

### LINKED ISSUE #262: Include BPTree root node cache size into memory used by indexes
`memory_used` function: https://github.com/qoollo/pearl/blob/master/src/blob/index/core.rs#L220
BPTree root node cache: https://github.com/qoollo/pearl/blob/master/src/blob/index/bptree/core.rs#L24

idruzhitskiy:
Closes #262


ikopylov (approved):
LGTM

PR #270, opened 2023-05-08
266 better btreemap size calculation

### LINKED ISSUE #266: Inaccurate calculation of the size of the BTreeMap in the index
The inaccuracy can be significant. And it might look like a memory leak to us

idruzhitskiy:
Closes #266

### DIFF
--- a/src/blob/index/core.rs
+++ b/src/blob/index/core.rs
@@ -67,6 +67,11 @@ where
     const BPTREE_ENTRY_SIZE: usize =
         size_of::<Vec<u8>>() + MemoryAttrs::<K>::KEY_SIZE + size_of::<Vec<RecordHeader>>();
     const RECORD_HEADER_SIZE: usize = size_of::<RecordHeader>() + MemoryAttrs::<K>::KEY_SIZE;
+    const BTREE_SIZE_MULTIPLIER: usize = (
+        size_of::<std::ptr::NonNull<()>>() + // btree node
+        size_of::<u16>() * 2 + // metadata
+        size_of::<std::ptr::NonNull<()>>() * 12 // edges vector in node
+    ) / 11; // count of keys in single node
 }
 
 pub type InMemoryIndex<K> = BTreeMap<K, Vec<RecordHeader>>;
@@ -104,7 +109,7 @@ where
         // last minus is neccessary, because allocated but not initialized record
         // headers don't have key allocated on heap
         MemoryAttrs::<K>::RECORD_HEADER_SIZE * records_allocated
-            + len * MemoryAttrs::<K>::BPTREE_ENTRY_SIZE
+            + len * (MemoryAttrs::<K>::BPTREE_ENTRY_SIZE + MemoryAttrs::<K>::BTREE_SIZE_MULTIPLIER)
             - (records_allocated - records_count) * MemoryAttrs::<K>::KEY_SIZE
     }
 
--- a/src/blob/index/core.rs
+++ b/src/blob/index/core.rs
@@ -64,7 +64,7 @@ where
     for<'a> K: Key<'a>,
 {
     const KEY_SIZE: usize = K::LEN as usize;
-    const BPTREE_ENTRY_SIZE: usize =
+    const BTREE_ENTRY_SIZE: usize =
         size_of::<Vec<u8>>() + MemoryAttrs::<K>::KEY_SIZE + size_of::<Vec<RecordHeader>>();
     const RECORD_HEADER_SIZE: usize = size_of::<RecordHeader>() + MemoryAttrs::<K>::KEY_SIZE;
     const BTREE_SIZE_MULTIPLIER: usize = (
@@ -105,11 +105,11 @@ where
         } = &mem;
         let len = self.headers.len();
         trace!("record_header_size: {}, records_allocated: {}, data.len(): {}, entry_size (key + vec): {}",
-                MemoryAttrs::<K>::RECORD_HEADER_SIZE, records_allocated, len, MemoryAttrs::<K>::BPTREE_ENTRY_SIZE);
+                MemoryAttrs::<K>::RECORD_HEADER_SIZE, records_allocated, len, MemoryAttrs::<K>::BTREE_ENTRY_SIZE);
         // last minus is neccessary, because allocated but not initialized record
         // headers don't have key allocated on heap
         MemoryAttrs::<K>::RECORD_HEADER_SIZE * records_allocated
-            + len * (MemoryAttrs::<K>::BPTREE_ENTRY_SIZE + MemoryAttrs::<K>::BTREE_SIZE_MULTIPLIER)
+            + len * (MemoryAttrs::<K>::BTREE_ENTRY_SIZE + MemoryAttrs::<K>::BTREE_SIZE_MULTIPLIER)
             - (records_allocated - records_count) * MemoryAttrs::<K>::KEY_SIZE
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Pearl changelog
 - Improved Arc placement in Storage (#185)
 - Improved visibility levels in Storage (#186)
 - BLOB-wide exclusive lock removed from the write path, instead the short lock on the in-memory index added (#249)
+- Include BTreeMap overhead in index calculations (#266)
 
 #### Fixed
 - Fix fork in test duplicating test environment (#254)


src/blob/index/core.rs
```diff
@@ -64,9 +64,18 @@ where
...
 {
     const KEY_SIZE: usize = K::LEN as usize;
-    const BPTREE_ENTRY_SIZE: usize =
+    const BTREE_ENTRY_SIZE: usize =
         size_of::<Vec<u8>>() + MemoryAttrs::<K>::KEY_SIZE + size_of::<Vec<RecordHeader>>();
     const RECORD_HEADER_SIZE: usize = size_of::<RecordHeader>() + MemoryAttrs::<K>::KEY_SIZE;
+    // Each node in BTreeMap contains preallocated vectors of 11 values and 12 edges.
+    // Although count of nodes can't be determined without reimplementing insertion algorithm,
+    // we can use approximation of overhead size added per one key
+    const BTREE_SIZE_MULTIPLIER: usize = (
```
ikopylov, line 73: This multiplier must be of type `f64`. Integer division will result in loss of precision. According to my calculations, the result of the division is `9.81`, which rounds up to `9` in integer division. So with 100_000 entries, this results in 81 KB of data being left out

src/blob/index/core.rs
```diff
@@ -64,9 +64,18 @@ where
...
+        size_of::<std::ptr::NonNull<()>>() +      // ptr to parent btree node
+        size_of::<u16>() * 2 +                    // metadata
+        size_of::<std::ptr::NonNull<()>>() * 12   // edges vector
```
ikopylov, line 76: According to the description in source code, the node is either `LeafNode` or `InternalNode`: https://doc.rust-lang.org/src/alloc/collections/btree/node.rs.html#123-129

`LeafNode` does not have edges. Only `InternalNode` have them. So the should be a separate calculation for `LeafNode` and for `InternalNode`

src/blob/index/core.rs
```diff
@@ -64,9 +64,18 @@ where
...
     const KEY_SIZE: usize = K::LEN as usize;
-    const BPTREE_ENTRY_SIZE: usize =
+    const BTREE_ENTRY_SIZE: usize =
         size_of::<Vec<u8>>() + MemoryAttrs::<K>::KEY_SIZE + size_of::<Vec<RecordHeader>>();
     const RECORD_HEADER_SIZE: usize = size_of::<RecordHeader>() + MemoryAttrs::<K>::KEY_SIZE;
+    // Each node in BTreeMap contains preallocated vectors of 11 values and 12 edges.
+    // Although count of nodes can't be determined without reimplementing insertion algorithm,
+    // we can use approximation of overhead size added per one key
+    const BTREE_SIZE_MULTIPLIER: usize = (
+        size_of::<std::ptr::NonNull<()>>() +      // ptr to parent btree node
```
ikopylov, line 74: According to the source code it is `Option<NonNull<...>>`. `Option` can add 1 byte on top. 
https://doc.rust-lang.org/src/alloc/collections/btree/node.rs.html#52

src/blob/index/core.rs
ikopylov: I think it would be good to define a `const fn` that calculates the multiplier and assign its result to a constant.

Also it would be good to declare constants: `BTREE_NODE_CAPACITY = 11` and `BTREE_NODE_EDGES_COUNT = 12` and use them in calculations

src/blob/index/core.rs
```diff
@@ -99,12 +108,12 @@ where
             ..
         } = &mem;
         let len = self.headers.len();
-        trace!("record_header_size: {}, records_allocated: {}, data.len(): {}, entry_size (key + vec): {}",
-                MemoryAttrs::<K>::RECORD_HEADER_SIZE, records_allocated, len, MemoryAttrs::<K>::BPTREE_ENTRY_SIZE);
+        trace!("record_header_size: {}, records_allocated: {}, data.len(): {}",
```
ikopylov, line 111: I think, there is no need to log the constant values. Perhaps, it will be better to log `len`, `records_allocated` and calculated memory after the calculation happened

src/blob/index/core.rs
ikopylov: Number of `LeafNode` can be estimated as `len / BTREE_NODE_CAPACITY`. `InternalNode`'s connects root with leafs, so it number can be estimated as `leaf_node_count / BTREE_NODE_EDGES_COUNT + leaf_node_count / BTREE_NODE_EDGES_COUNT^2 + ... + leaf_node_count / BTREE_NODE_EDGES_COUNT^5`. I think it would be enough to calculate it up to the 5th power

src/blob/index/core.rs
```diff
@@ -59,14 +59,29 @@ struct MemoryAttrs<K> {
...
 {
     const KEY_SIZE: usize = K::LEN as usize;
-    const BPTREE_ENTRY_SIZE: usize =
+    const BTREE_ENTRY_SIZE: usize =
         size_of::<Vec<u8>>() + MemoryAttrs::<K>::KEY_SIZE + size_of::<Vec<RecordHeader>>();
     const RECORD_HEADER_SIZE: usize = size_of::<RecordHeader>() + MemoryAttrs::<K>::KEY_SIZE;
+    // Each node in BTreeMap contains preallocated vectors of 11 values and 12 edges.
+    // Although count of nodes can't be determined without reimplementing insertion algorithm,
+    // we can use approximation of overhead size added per one key
+    const BTREE_SIZE_MULTIPLIER: f64 =
```
ikopylov, line 77: I suggest to split this into 3 const:
1. `BTREE_LEAF_NODE_SIZE = ...`
2. `BTREE_INTERNAL_NODE_SIZE = BTREE_LEAF_NODE_SIZE + ...`
3. `BTREE_SIZE_MULTIPLIER = ...`

src/blob/index/core.rs
```diff
@@ -59,14 +59,29 @@ struct MemoryAttrs<K> {
...
     const RECORD_HEADER_SIZE: usize = size_of::<RecordHeader>() + MemoryAttrs::<K>::KEY_SIZE;
+    // Each node in BTreeMap contains preallocated vectors of 11 values and 12 edges.
+    // Although count of nodes can't be determined without reimplementing insertion algorithm,
+    // we can use approximation of overhead size added per one key
+    const BTREE_SIZE_MULTIPLIER: f64 = 
+        ((size_of::<Option<std::ptr::NonNull<()>>>() +                     // ptr to parent
+          size_of::<u16>() * 2 +                                           // metadata
+          MemoryAttrs::<K>::BTREE_ENTRY_SIZE * BTREE_VALUES_LEN) as f64    // vector of values
+         / BTREE_VALUES_LEN as f64) +                                      // data exists every `BTREE_VALUES_LEN` nodes
+        ((size_of::<std::ptr::NonNull<()>>() * BTREE_EDGES_LEN) as f64 /   // vector of edges
```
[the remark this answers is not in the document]
ikopylov: Also there should be a multiplication (not division):
```
BTREE_INTERNAL_NODE_SIZE * (1 / BTREE_VALUES_LEN / BTREE_EDGES_LEN + 1 / BTREE_VALUES_LEN / BTREE_EDGES_LEN^2 + ...  + 1 / BTREE_VALUES_LEN / BTREE_EDGES_LEN^5) = 
= BTREE_INTERNAL_NODE_SIZE * (1 + BTREE_EDGES_LEN + BTREE_EDGES_LEN^2 + ... + BTREE_EDGES_LEN^5) / BTREE_VALUES_LEN =
= BTREE_INTERNAL_NODE_SIZE  * (1 + BTREE_EDGES_LEN * (1 + BTREE_EDGES_LEN * (1 + BTREE_EDGES_LEN  * (1 + BTREE_EDGES_LEN * (1 + BTREE_EDGES_LEN))))) / BTREE_VALUES_LEN 
```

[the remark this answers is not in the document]
ikopylov: Sorry, I lost the division by `BTREE_EDGES_LEN^5` in formula. It should be:
```
= BTREE_INTERNAL_NODE_SIZE  * (1 + BTREE_EDGES_LEN * (1 + BTREE_EDGES_LEN * (1 + BTREE_EDGES_LEN  * (1 + BTREE_EDGES_LEN * (1 + BTREE_EDGES_LEN))))) / (BTREE_VALUES_LEN * BTREE_EDGES_LEN^5)
```

src/blob/index/core.rs
```diff
@@ -59,14 +59,35 @@ struct MemoryAttrs<K> {
...
+                                        size_of::<u16>() * 2 +                                           // metadata
+                                        MemoryAttrs::<K>::BTREE_ENTRY_SIZE * BTREE_VALUES_LEN;           // data
+    const BTREE_DATA_NODE_RATIO: f64 = 1.0 / BTREE_VALUES_LEN as f64;
+    const BTREE_INTERNAL_NODE_OVERHEAD: usize = size_of::<std::ptr::NonNull<()>>() * BTREE_EDGES_LEN;    // edges
+    const BTREE_INTERNAL_NODE_RATIO: f64 = (((1.0 +
+                                              BTREE_EDGES_LEN as f64 +
+                                              BTREE_EDGES_LEN.pow(2) as f64 +
+                                              BTREE_EDGES_LEN.pow(3) as f64 +
+                                              BTREE_EDGES_LEN.pow(4) as f64)
+                                             / (BTREE_VALUES_LEN as f64 * BTREE_EDGES_LEN.pow(5) as f64)));
```
ikopylov, line 87: You can do the type conversion once for the numerator and once for the denominator

src/blob/index/core.rs
```diff
@@ -59,14 +59,34 @@ struct MemoryAttrs<K> {
...
 impl<K> MemoryAttrs<K>
 where
     for<'a> K: Key<'a>,
 {
-    const KEY_SIZE: usize = K::LEN as usize;
-    const BPTREE_ENTRY_SIZE: usize =
-        size_of::<Vec<u8>>() + MemoryAttrs::<K>::KEY_SIZE + size_of::<Vec<RecordHeader>>();
-    const RECORD_HEADER_SIZE: usize = size_of::<RecordHeader>() + MemoryAttrs::<K>::KEY_SIZE;
+    const BTREE_ENTRY_SIZE: usize =
+        size_of::<Vec<u8>>() + K::MEM_SIZE + size_of::<Vec<RecordHeader>>();
```
ikopylov, line 71: `size_of::<Vec<u8>>()` should be removed due to `K::MEM_SIZE` usage

src/blob/index/core.rs
```diff
@@ -59,14 +59,34 @@ struct MemoryAttrs<K> {
...
+    const RECORD_HEADER_SIZE: usize = size_of::<RecordHeader>() + K::MEM_SIZE;
```
ikopylov, line 72: Here, the previous implementation was correct. It should be `size_of::<RecordHeader>() + MemoryAttrs::<K>::KEY_SIZE`, because `RecordHeader` stores key inside a vector

tests/tests.rs
```diff
@@ -906,9 +906,7 @@ async fn test_memory_index() {
     // Key and data size - size of entries in binarymap not including size of entry internal value (RecordHeader)
     // RecordHeader is private, so instead of measurement size_of::Vec<RecordHeader>() there is
     // measurement size_of::<Vec<u8>>(), which has the same size on stack
-    const KEY_AND_DATA_SIZE: usize = std::mem::size_of::<Vec<u8>>()
-        + std::mem::size_of::<u32>()
-        + std::mem::size_of::<Vec<u8>>();
+    const KEY_AND_DATA_SIZE: usize = KeyTest::MEM_SIZE;
```
ikopylov, line 909: You lost one `std::mem::size_of::<Vec<u8>>()`

tests/tests.rs
```diff
@@ -922,8 +920,11 @@ async fn test_memory_index() {
     // rule higher). Then there is the amortized growth.
     assert_eq!(
         storage.index_memory().await,
-        KEY_AND_DATA_SIZE * 4 + RECORD_HEADER_SIZE * 13 - 6 * std::mem::size_of::<u32>()
-    ); // 4 keys, 7 records in active blob (13 allocated (6 without key on heap))
+        KEY_AND_DATA_SIZE * 4 + // 4 keys
+        RECORD_HEADER_SIZE * 13 - // 13 allocated
+        6 * KeyTest::MEM_SIZE + // 6 without key on heap
+        103// btree overhead
```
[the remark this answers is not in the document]
idruzhitskiy: It's a lot harder since we have data and key taken out here (in index calc they are insed btree size mutiplier). I'd prefer to leave numbers here

ikopylov (approved):
LGTM

PR #280, opened 2023-05-18
274 entry load data checksum validation

### LINKED ISSUE #274: `Entry::load_data` should validate checksum
https://github.com/qoollo/pearl/blob/master/src/blob/entry.rs#L45

ikopylov:
Closes #274


idruzhitskiy (approved):
LGTM

PR #281, opened 2023-05-19
260 reduce internal struct mem size

### LINKED ISSUE #260: Reduce internal strucutre size in memory
There are several places where we can reduce the size of internal structs. Since there are expected to be about 100k instances of `Storage` in production, even a small reduction can give good improvements in memory consumption.

1. Bloom filter config sharing. It is expected, that all instances will have the same config. This shared value can be stored in the static field with lazy initialization (https://github.com/qoollo/pearl/blob/master/src/filter/bloom.rs#L16)
2. `FileName` consumes too much memory (https://github.com/qoollo/pearl/blob/master/src/blob/core.rs#L466)

ikopylov:
Closes #260

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ Pearl changelog
 
 
 #### Changed
-
+- Reduced internal strucutre size in memory: bloom filter `Config` and `FileName` (#260)
 
 #### Fixed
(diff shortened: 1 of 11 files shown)


idruzhitskiy (approved):
LGTM

PR #285, opened 2023-05-30
284 delete closed only if exist

### LINKED ISSUE #284: Deletion record for old BLOBs should only be written if the key is present
Right now there is a bug and a deletion record is added to every old BLOB:
https://github.com/qoollo/pearl/blob/master/src/storage/core.rs#L1034

Related to https://github.com/qoollo/bob/issues/797

idruzhitskiy:
Closes #284

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ Pearl changelog
 
 #### Fixed
 - Use `dep:<crate_name>` syntax in feature declaration to avoid unnecessary feature flags (#272) 
-- Only add deletion record in old blobs if key is presented (#284)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ Pearl changelog
 
 #### Fixed
 - Use `dep:<crate_name>` syntax in feature declaration to avoid unnecessary feature flags (#272) 
+- Only add deletion record in old blobs if key is presented (#284)
```
ikopylov, line 14:
```suggestion
- Only add a deletion record in old blobs if the key is present (#284)
```

PR #288, opened 2023-06-05
234 auto flush

### LINKED ISSUE #234: Significant `put` performance drop due to `fsync_all` call on blob close

### LINKED ISSUE #243: Do `fsync` after `write_header` in BLOB
https://github.com/qoollo/pearl/blob/master/src/blob/core.rs#L80

idruzhitskiy:
Closes #234
Closes #243

### DIFF
--- a/src/blob/core.rs
+++ b/src/blob/core.rs
@@ -78,6 +78,7 @@ where
         let mut buf = BytesMut::with_capacity(size as usize);
         serialize_into((&mut buf).writer(), &self.header)?;
         self.file.write_append_all(buf.freeze()).await?;
+        self.file.fsyncdata().await?;
         Ok(())
     }
 
--- a/src/blob/core.rs
+++ b/src/blob/core.rs
@@ -30,6 +30,16 @@ where
     validate_data_during_index_regen: bool,
 }
 
+pub(crate) struct WriteResult {
+    dirty_bytes: u64
+}
+
+impl WriteResult {
+    pub(crate) fn dirty_bytes(&self) -> u64 {
+        self.dirty_bytes
+    }
+}
+
 impl<K> Blob<K>
 where
     for<'a> K: Key<'a> + 'static,
@@ -242,7 +252,7 @@ where
         // @TODO implement
     }
 
-    pub(crate) async fn write(blob: &ASRwLock<Self>, key: &K, record: Record) -> Result<()> {
+    pub(crate) async fn write(blob: &ASRwLock<Self>, key: &K, record: Record) -> Result<WriteResult> {
         debug!("blob write");
         let (partially_serialized, mut header) = record.to_partially_serialized_and_header()?;
         // Only one upgradable_read lock is allowed at a time. This is critical because we want to
@@ -251,7 +261,7 @@ where
         let write_result = partially_serialized.write_to_file(&blob.file).await?;
         header.set_offset_checksum(write_result.blob_offset(), write_result.header_checksum());
         blob.index.push(key, header)?;
-        Ok(())
+        Ok(WriteResult { dirty_bytes: blob.file.dirty_bytes() })
     }
 
     async fn write_mut(&mut self, key: &K, record: Record) -> Result<RecordHeader> {
--- a/src/blob/core.rs
+++ b/src/blob/core.rs
@@ -31,28 +31,12 @@ where
 }
 
 pub(crate) struct WriteResult {
-    dirty_bytes: u64
-}
-
-impl WriteResult {
-    pub(crate) fn dirty_bytes(&self) -> u64 {
-        self.dirty_bytes
-    }
+    pub dirty_bytes: u64
 }
 
 pub(crate) struct DeleteResult {
-    dirty_bytes: u64,
-    success: bool
-}
-
-impl DeleteResult {
-    pub(crate) fn dirty_bytes(&self) -> u64 {
-        self.dirty_bytes
-    }
-
-    pub(crate) fn success(&self) -> bool {
-        self.success
-    }
+    pub dirty_bytes: u64,
+    pub deleted: bool
 }
 
 impl<K> Blob<K>
@@ -279,13 +263,13 @@ where
         Ok(WriteResult { dirty_bytes: blob.file.dirty_bytes() })
     }
 
-    async fn write_mut(&mut self, key: &K, record: Record) -> Result<RecordHeader> {
+    async fn write_mut(&mut self, key: &K, record: Record) -> Result<WriteResult> {
         debug!("blob write");
         let (record, mut header) = record.to_partially_serialized_and_header()?;
         let write_result = record.write_to_file(&self.file).await?;
         header.set_offset_checksum(write_result.blob_offset(), write_result.header_checksum());
-        self.index.push(key, header.clone())?;
-        Ok(header)
+        self.index.push(key, header)?;
+        Ok(WriteResult { dirty_bytes: self.file.dirty_bytes() })
     }
 
     pub(crate) async fn read_last(
@@ -351,7 +335,7 @@ where
         if !only_if_presented || self.index.get_any(key).await?.is_found() {
             self.push_deletion_record(key).await
         } else {
-            Ok(DeleteResult { dirty_bytes: self.file.dirty_bytes(), success: false })
+            Ok(DeleteResult { dirty_bytes: self.file.dirty_bytes(), deleted: false })
         }
     }
 
@@ -361,10 +345,8 @@ where
             self.load_index().await?;
         }
         let record = Record::deleted(key)?;
-        let header = self.write_mut(key, record).await?;
-        self.index.push_deletion(key, header)?;
-        let dirty_bytes = self.file.dirty_bytes();
-        Ok(DeleteResult { dirty_bytes, success: true })
+        let result = self.write_mut(key, record).await?;
+        Ok(DeleteResult { dirty_bytes: result.dirty_bytes, deleted: true })
     }
 
     fn headers_to_entries(headers: Vec<RecordHeader>, file: &File) -> Vec<Entry> {
--- a/src/blob/core.rs
+++ b/src/blob/core.rs
@@ -40,6 +40,21 @@ impl WriteResult {
     }
 }
 
+pub(crate) struct DeleteResult {
+    dirty_bytes: u64,
+    success: bool
+}
+
+impl DeleteResult {
+    pub(crate) fn dirty_bytes(&self) -> u64 {
+        self.dirty_bytes
+    }
+
+    pub(crate) fn success(&self) -> bool {
+        self.success
+    }
+}
+
 impl<K> Blob<K>
 where
     for<'a> K: Key<'a> + 'static,
@@ -332,23 +347,24 @@ where
         &mut self,
         key: &K,
         only_if_presented: bool,
-    ) -> Result<bool> {
+    ) -> Result<DeleteResult> {
         if !only_if_presented || self.index.get_any(key).await?.is_found() {
-            self.push_deletion_record(key).await?;
-            Ok(true)
+            self.push_deletion_record(key).await
         } else {
-            Ok(false)
+            Ok(DeleteResult { dirty_bytes: self.file.dirty_bytes(), success: false })
         }
     }
 
-    async fn push_deletion_record(&mut self, key: &K) -> Result<()> {
+    async fn push_deletion_record(&mut self, key: &K) -> Result<DeleteResult> {
         let on_disk = self.index.on_disk();
         if on_disk {
             self.load_index().await?;
         }
         let record = Record::deleted(key)?;
         let header = self.write_mut(key, record).await?;
-        self.index.push_deletion(key, header)
+        self.index.push_deletion(key, header)?;
+        let dirty_bytes = self.file.dirty_bytes();
+        Ok(DeleteResult { dirty_bytes, success: true })
     }
 
     fn headers_to_entries(headers: Vec<RecordHeader>, file: &File) -> Vec<Entry> {
@@ -438,6 +454,10 @@ where
 
     pub(crate) fn records_count(&self) -> usize {
         self.index.count()
+    }
+
+    pub(crate) fn file_dirty_bytes(&self) -> u64 {
+        self.file.dirty_bytes()
     }
 
     pub(crate) async fn fsyncdata(&self) -> IOResult<()> {
--- a/src/io/unix/iouring.rs
+++ b/src/io/unix/iouring.rs
@@ -139,7 +139,10 @@ impl File {
 
     pub(crate) async fn fsyncdata(&self) -> IOResult<()> {
         if let Some(ref rio) = self.rio {
-            rio.fsync(self.sync.std_file_ref()).await
+            let size = self.size();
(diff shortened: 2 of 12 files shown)


src/io/unix/iouring.rs
```diff
@@ -69,6 +69,10 @@ impl File {
         self.sync.size()
     }
 
+    pub(crate) fn dirty_bytes(&self) -> u64 {
```
ikopylov, line 72: `fsync` in this file should be updated too, when `rio` is used

src/io/windows/sync.rs
```diff
@@ -44,14 +44,21 @@ pub(crate) struct File {
...
+    size: AtomicU64,
+    synced_size: AtomicU64
 }
 
 
 impl File {
     pub(crate) fn size(&self) -> u64 {
         self.inner.size.load(Ordering::SeqCst)
     }
+    pub(crate) fn synced_size(&self) -> u64 {
```
ikopylov, line 56: `fsync` in this file should update `synced_size`

src/io/unix/sync.rs
```diff
@@ -156,7 +163,10 @@ impl File {
 
     pub(crate) async fn fsyncdata(&self) -> IOResult<()> {
         let file_inner = self.inner.clone();
-        Self::background_sync_call(move || file_inner.std_file.sync_all()).await
+        let size = self.size();
+        Self::background_sync_call(
+            move || file_inner.std_file.sync_all().map(|_| file_inner.synced_size.store(size, Ordering::SeqCst))
```
ikopylov, line 168: I'd prefer:
```suggestion
            move || {
               file_inner.std_file.sync_all()?;
               file_inner.synced_size.store(size, Ordering::SeqCst);
               Ok(())
            }
```

src/storage/config.rs
```diff
@@ -175,6 +184,7 @@ impl Default for Config {
             deferred_min_time: Duration::from_secs(60),
             deferred_max_time: Duration::from_secs(180),
             validate_data_during_index_regen: false,
+            max_dirty_bytes_before_sync: 1024 * 1024
```
ikopylov, line 187: 32Mb

src/storage/core.rs
```diff
@@ -41,6 +41,7 @@ where
 {
     inner: Arc<Inner<K>>,
     observer: Observer<K>,
+    max_dirty_bytes_before_sync: u64
```
ikopylov, line 44: There is no need to store that field. Value is available from config stored within `inner`

src/blob/core.rs
```diff
@@ -250,7 +261,7 @@ where
         let write_result = partially_serialized.write_to_file(&blob.file).await?;
         header.set_offset_checksum(write_result.blob_offset(), write_result.header_checksum());
         blob.index.push(key, header)?;
-        Ok(())
+        Ok(WriteResult { dirty_bytes: blob.file.dirty_bytes() })
```
ikopylov, line 264: I think you can expose `dirty_bytes` as a function inside `Blob` structure. In this way there is no need to update result type of `write` function

src/storage/core.rs
```diff
@@ -335,9 +337,12 @@ where
             } else {
                 Err(e.into())
             }
-        });
+        })?;
         self.try_update_active_blob(blob).await?;
-        result
+        if result.dirty_bytes() > self.max_dirty_bytes_before_sync {
```
ikopylov, line 342: Same should be done for `delete`

src/storage/observer_worker.rs
```diff
@@ -226,6 +235,24 @@ where
         return true;
     }
 
+    async fn try_run_fsync_task(&mut self) -> bool {
+        if self.fsync_task.as_ref().map_or(false, |task| !task.is_finished()) {
+            // Task is in progress. Avoid starting second one
+            return false;
+        }
+
+        complete_task(&mut self.fsync_task, "index_dump_task").await;
```
ikopylov, line 244:
```suggestion
        complete_task(&mut self.fsync_task, "fsync_task").await;

```

src/storage/core.rs
```diff
@@ -1228,7 +1233,7 @@ where
         }
     }
 
-    async fn fsyncdata(&self) -> IOResult<()> {
+    pub(crate) async fn fsyncdata(&self) -> IOResult<()> {
         self.safe.read().await.fsyncdata().await
```
ikopylov, line 1237: Additional check, that `dirty_bytes() > max_dirty_bytes_before_sync` is needed here to prevent unneeded call to `fsync`

src/storage/observer.rs
```diff
@@ -146,6 +145,11 @@ where
             .await
     }
 
+    pub(crate) async fn try_fsync_data(&self) {
+        self.send_msg(Msg::new(OperationType::TryFsyncData, None))
```
ikopylov, line 149: There is a problem here: after the `dirty_bytes` exceeds the threshold, messages will be sent on each new record, which can lead to queue overflow. An additional message frequency limit is required. 
One possible approach is to store the last time the message was sent and ignore sending new once if some constant amount of time was not passed (e.g. 500ms). Time measurement can be done with `coarsetime` crate. Time can be stored in `AtomicU64` as milliseconds elapsed from start of the epoch (Unix time).
Another approach is to track, that `fsync` is in progress with `AtomicBool` inside `Inner<K>` struct and don't send new messages if it is in progress.

src/blob/core.rs
idruzhitskiy: We acquire readlock here, and if we expose it as a function we'll need to acquire another readlock on write

src/storage/observer.rs
idruzhitskiy: I agree, I'll implement the latter aproach

src/blob/core.rs
```diff
@@ -30,6 +30,31 @@ where
     validate_data_during_index_regen: bool,
 }
 
+pub(crate) struct WriteResult {
+    dirty_bytes: u64
+}
+
+impl WriteResult {
```
ikopylov, line 37: Since it's only an internal structure, I think it's ok to declare its fields with `pub` modifier. This will remove the `impl` altogether and reduce the size of the code a bit.

The same is applicable to `DeleteResult`

src/blob/core.rs
```diff
@@ -30,6 +30,31 @@ where
...
+    pub(crate) fn dirty_bytes(&self) -> u64 {
+        self.dirty_bytes
+    }
+}
+
+pub(crate) struct DeleteResult {
+    dirty_bytes: u64,
+    success: bool
```
ikopylov, line 45: According to the code below, this field means that the record was present and therefore deleted. I wouldn't call the field like that in that case. The better name would be `deleted`

src/blob/core.rs
```diff
@@ -321,23 +347,24 @@ where
...
-    async fn push_deletion_record(&mut self, key: &K) -> Result<()> {
+    async fn push_deletion_record(&mut self, key: &K) -> Result<DeleteResult> {
         let on_disk = self.index.on_disk();
         if on_disk {
             self.load_index().await?;
         }
         let record = Record::deleted(key)?;
         let header = self.write_mut(key, record).await?;
-        self.index.push_deletion(key, header)
+        self.index.push_deletion(key, header)?;
```
ikopylov, line 365: Just noticed a bug. `write_mut` already adds the header to `index`. This addition is the second one. So this results in 2 identical headers inside the index

src/storage/config.rs
```diff
@@ -175,6 +184,7 @@ impl Default for Config {
             deferred_min_time: Duration::from_secs(60),
             deferred_max_time: Duration::from_secs(180),
             validate_data_during_index_regen: false,
+            max_dirty_bytes_before_sync: 1//32 * 1024 * 1024
```
ikopylov, line 187:
```suggestion
            max_dirty_bytes_before_sync: 32 * 1024 * 1024
```

src/storage/core.rs
```diff
@@ -1016,12 +1022,18 @@ where
...
-        let deleted_in_active = Self::mark_all_as_deleted_active(safe, key, only_if_presented).await?;
+        let deleted_in_active_result = Self::mark_all_as_deleted_active(safe, key, only_if_presented).await?;
+        let deleted_in_active = deleted_in_active_result.as_ref().map(|r| if r.success() { 1 } else { 0 }).unwrap_or(0);
         let deleted_in_closed = Self::mark_all_as_deleted_closed(safe, key).await?;
 
         if deleted_in_closed > 0 {
             self.observer.defer_dump_old_blob_indexes().await;
         }
+        if let Some(result) = deleted_in_active_result {
+            if self.inner.should_try_fsync(result.dirty_bytes()) {
```
ikopylov, line 1033: I think this can be done inside `mark_all_as_deleted_active`. In this way the code changes will be within the single function

src/storage/core.rs
```diff
@@ -1228,14 +1241,31 @@ where
...
-        self.safe.read().await.fsyncdata().await
+    pub(crate) async fn fsyncdata(&self) -> IOResult<()> {
+        let safe = self.safe.read().await;
+        if let Some(ablob) = &safe.active_blob {
+            let ablob = ablob.read().await;
+            if !self.should_try_fsync(ablob.file_dirty_bytes()) {
+                return Ok(());
+            }
+        }
+        self.fsync_in_progress.store(true, Ordering::Release);
```
ikopylov, line 1252: There are 2 problems here:
1. Right now there is no guarantee that `fsync_in_progress` will be reset to `false`, because of `?` on the line below. The best way to solve the problem is to declare a helper struct to use RAII feature. This struct will hold a reference to `AtomicBool` and resets its value to `false` on drop
2. It is better to set `fsync_in_progress` to `true` as early as possible to prevent large number of messages in the `Observer` queue. At the beginning of the function `read` lock is acquired and that can took some time. Thus it is better to set  `fsync_in_progress` to `true` before that (right at the beginning of this function)

src/storage/core.rs
idruzhitskiy: We don't have inner or observer inside `mark_all` function, so it can't be moved here without adding 2 more args to the function call

src/storage/core.rs
```diff
@@ -1228,14 +1241,50 @@ where
...
     }
 
-    async fn fsyncdata(&self) -> IOResult<()> {
-        self.safe.read().await.fsyncdata().await
+    pub(crate) async fn fsyncdata(&self) -> IOResult<()> {
+        if self.fsync_in_progress.load(Ordering::Acquire) {
+            return Ok(())
+        }
+
+        self.fsync_in_progress.store(true, Ordering::Release);
```
ikopylov, line 1249: It is better to use `compare_exchange` instead of `load` -> `store` functions

src/storage/core.rs
```diff
@@ -1228,14 +1241,50 @@ where
...
     /// Dumps indexes on old blobs. This method is slow, so it is better to run it in background
     pub(crate) async fn try_dump_old_blob_indexes(&self) {
         Safe::try_dump_old_blob_indexes(&self.safe, self.get_dump_sem(), Duration::from_millis(200)).await;
     }
+
+    pub(crate) fn should_try_fsync(&self, dirty_bytes: u64) -> bool {
+        if self.too_many_dirty_bytes(dirty_bytes) {
+            return !self.fsync_in_progress.load(Ordering::Acquire);
+        }
+        false
```
ikopylov, line 1272:
```suggestion
        self.too_many_dirty_bytes(dirty_bytes) && !self.fsync_in_progress.load(Ordering::Acquire)
```

src/storage/core.rs
```diff
@@ -1057,24 +1069,24 @@ where
         safe: &Safe<K>,
         key: &K,
         only_if_presented: bool,
-    ) -> Result<u64> {
+    ) -> Result<Option<DeleteResult>> {
```
ikopylov, line 1072: Looks like `Option` is redundant here. If there is no active BLOB, then `DeleteResult { dirty_bytes: 0, deleted: false }` can be returned. This way the code in `delete_core` will be a bit simpler

src/storage/core.rs
```diff
@@ -1057,24 +1069,24 @@ where
...
                 .write()
                 .await
                 .mark_all_as_deleted(key, only_if_presented)
                 .await?;
-            let count = if is_deleted { 1 } else { 0 };
-            debug!("{} deleted from active blob", count);
-            count
+            if result.deleted {
+                debug!("Deleted record from active blob");
+            }
```
ikopylov, line 1085:
```suggestion
            debug!("Deleted {} records from active blob", if result.deleted { 1 } else { 0 });
```

src/storage/core.rs
```diff
@@ -1228,14 +1239,46 @@ where
         }
     }
 
-    async fn fsyncdata(&self) -> IOResult<()> {
-        self.safe.read().await.fsyncdata().await
+    pub(crate) async fn fsyncdata(&self) -> IOResult<()> {
+        if self.fsync_in_progress.compare_exchange(false, true, Ordering::Release, Ordering::Acquire).is_err() {
```
ikopylov, line 1243: According to the documentation, first ordering is used both for load and store, so it should be `AcqRel`

```suggestion
        if self.fsync_in_progress.compare_exchange(false, true, Ordering::AcqRel, Ordering::Acquire).is_err() {
```

src/io/unix/sync.rs
```diff
@@ -156,7 +163,14 @@ impl File {
 
     pub(crate) async fn fsyncdata(&self) -> IOResult<()> {
         let file_inner = self.inner.clone();
-        Self::background_sync_call(move || file_inner.std_file.sync_all()).await
+        let size = self.size();
+        Self::background_sync_call(
+            move || {
+               file_inner.std_file.sync_all()?;
+               file_inner.synced_size.store(size, Ordering::SeqCst);
```
ikopylov, line 170: I suggest to use `fetch_max` here for safety
```suggestion
               file_inner.synced_size.fetch_max(size, Ordering::SeqCst);
```

src/io/unix/sync.rs
```diff
@@ -174,6 +188,11 @@ impl File {
         self.inner.size.fetch_add(len, Ordering::SeqCst)
     }
 
+    #[cfg(feature = "async-io-rio")]
+    pub(super) fn set_synced_size(&self, synced_size: u64) {
+        self.inner.synced_size.store(synced_size, Ordering::SeqCst);
```
ikopylov, line 193: Same here 
```suggestion
        self.inner.synced_size.fetch_max(synced_size, Ordering::SeqCst);
```

src/io/windows/sync.rs
```diff
@@ -113,7 +120,14 @@ impl File {
 
     pub(crate) async fn fsyncdata(&self) -> IOResult<()> {
         let file_inner = self.inner.clone();
-        Self::background_sync_call(move || file_inner.std_file.sync_all()).await
+        let size = self.size();
+        Self::background_sync_call(
+            move || {
+               file_inner.std_file.sync_all()?;
+               file_inner.synced_size.store(size, Ordering::SeqCst);
```
ikopylov, line 127: Same here
```suggestion
               file_inner.synced_size.fetch_max(size, Ordering::SeqCst);
```

ikopylov (approved):
LGTM

